PR #14, opened 2026-05-23
Inline vault_path in commit-capture hook output

### LINKED ISSUE #12: commit-capture hook re-reads obsidian.local.md on every commit
## Problem

The commit-capture PostToolUse hook resolves the plugin dir and reads `obsidian.local.md` on every successful `git commit` to fetch the vault path and routing rules. Across an active day this is ~1-2K tokens per commit of redundant config reads.

## Proposal

Cache the parsed config (vault_path, domain routing) once per session, or have the hook write a resolved-path file the script can stat-check before re-parsing. Routing rules change rarely; per-commit re-parse is unnecessary.

## Notes

- Already partially mitigated by caching routing in `~/.claude/projects/.../memory/MEMORY.md`, but that's user-side, not hook-side.
- Plugin version bumps regularly so the resolver itself (`ls -1d .../obsidian/*/ | sort -V | tail -1`) should stay dynamic — only the config read needs caching.

nhangen:
Closes #12

## Description

The `obsidian-commit-capture` skill previously had the agent `Read` `${CLAUDE_PLUGIN_ROOT}/obsidian.local.md` on every successful git commit to resolve the vault path. Across an active day this is ~1–2K tokens per commit of redundant config reads.

`scripts/commit-capture.sh` now parses `vault_path` from the YAML frontmatter once per commit (shell parse, zero token cost) and appends it to the inline metadata line. `SKILL.md` is updated to consume `vault_path` from the hook output and explicitly forbid reading the config file from the skill.

If the config is missing or unparseable, `vault_path` is emitted empty and the skill skips silently — same behavior as before.

## Testing Procedure

1. `git checkout nh/perf/12-cache-config`
2. Mock the config:
   ```bash
   mkdir -p /tmp/obs-test && printf -- '---\nvault_path: ~/Documents/Obsidian\n---\n' > /tmp/obs-test/obsidian.local.md
   INPUT='{"tool_input":{"command":"git commit -m foo"},"tool_response":{"stdout":"[main abc1234] foo\n"}}'
   CLAUDE_PLUGIN_ROOT=/tmp/obs-test bash scripts/commit-capture.sh <<< "$INPUT"
   ```
3. Confirm the output line includes `| vault_path=/Users/<you>/Documents/Obsidian`.
4. Run again with `CLAUDE_PLUGIN_ROOT=/tmp/does-not-exist` and confirm `vault_path=` (empty) is emitted — the skill will skip silently.

## Additional Notes

Followups left out of scope: cache-by-mtime if the YAML parse ever becomes the bottleneck. Currently the parse is a single while-read pass over a small file, negligible compared to the per-commit git commands the hook already runs.

### DIFF
--- a/scripts/commit-capture.sh
+++ b/scripts/commit-capture.sh
@@ -110,7 +110,17 @@ if [ -n "${CLAUDE_PLUGIN_ROOT:-}" ] && [ -f "$CONFIG_FILE" ]; then
     case "$line" in
       vault_path:*)
         VAULT_PATH="${line#vault_path:}"
+        # Strip trailing CR from CRLF-edited configs before any other parsing.
+        VAULT_PATH="${VAULT_PATH%$'\r'}"
+        # Strip an inline "# comment" tail before quote/whitespace handling.
+        case "$VAULT_PATH" in
+          *' #'*) VAULT_PATH="${VAULT_PATH%% #*}" ;;
+          *$'\t#'*) VAULT_PATH="${VAULT_PATH%%	#*}" ;;
+        esac
         VAULT_PATH="${VAULT_PATH# }"
+        VAULT_PATH="${VAULT_PATH#	}"
+        VAULT_PATH="${VAULT_PATH% }"
+        VAULT_PATH="${VAULT_PATH%	}"
         VAULT_PATH="${VAULT_PATH#\"}"
         VAULT_PATH="${VAULT_PATH%\"}"
         VAULT_PATH="${VAULT_PATH#\'}"
--- a/scripts/commit-capture.sh
+++ b/scripts/commit-capture.sh
@@ -98,7 +98,34 @@ while [ -n "$REMAINING" ]; do
   esac
 done
 
+# --- Resolve vault_path from obsidian.local.md ---
+# Inline this in the hook output so the skill doesn't have to Read the config
+# file on every commit. The config rarely changes; if it does, the next commit
+# picks up the new value.
+
+VAULT_PATH=""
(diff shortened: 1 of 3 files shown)


nhangen (commented):
## Self-review summary (pr-review-panel mini)

Skeptic auditor collapsed two HIGH silent-failure findings on inspection: empty `CLAUDE_PLUGIN_ROOT` skip and ` | ` delimiter collision are both **pre-existing** in commit-capture.sh — this PR doesn't introduce or worsen them. Headline net-new defect is YAML sanitization.

### Findings

**MEDIUM — YAML parser: no `\r` / BOM / inline-`#` / multiline / `~user/` handling** (scripts/commit-capture.sh:109-125)
Real corruption modes: trailing `\r` from CRLF configs, `vault_path: ~/foo  # comment` swallows the comment into the path, `~user/Documents` passes through unexpanded. Fix: strip `\r` + trailing `# ...` after parsing.

**MEDIUM — No automated test for the parse block** (tests/)
Repo has `tests/session-summarize-intent.sh` as a precedent harness. Reverting the parse block today produces zero test failures.

**LOW — `vault_path:*` matches anywhere in file** (scripts/commit-capture.sh:111)
No `---` fence tracking. First match wins regardless of context.

**LOW — `${HOME}` unguarded in tilde expansion** (scripts/commit-capture.sh:119)
Per `shell-required-env-vars`: hook can run under stripped env where HOME may be empty.

### Plan
Fixing MEDIUM #1 (YAML sanitization) and MEDIUM #2 (add `tests/commit-capture-vault-path.sh`) in follow-up commits.

<sub>— reviewer (pr-review-panel mini)</sub>

Issue #26, opened 2026-06-30
Vaultkeeper "sole brain": move routing/dedup/MOC judgment into the keeper


nhangen:
## Deferred from the "sole writer now, brain later" arc (v1.11.0–v1.15.0)

The keeper is now the **single writer**: the `keeper` CLI (`append` + `insert`) is the one deterministic write implementation, and `keeper-save` routes mechanical/resolved writes to it while judgment writes (routing, dedup, MOC promotion) still go through the `vault-librarian` agent.

What's **not** yet done — the "brain" half: the routing/dedup/MOC-promotion *cognition* still lives in the skills (`save-conversation` infers intent, runs the cross-domain tiebreaker, same-day dedup, MOC-promotion prompt) and in the librarian agent. The deferred step is moving that judgment **into** the keeper so callers become thin "build payload + hand off," and there is one place that decides where a note goes and whether it's a duplicate.

### Scope (its own design+audit cycle when picked up)
- Decide what "routing/dedup brain in the keeper" means concretely: a richer `keeper route`/`keeper file` surface vs. the librarian agent staying the brain but becoming the sole judgment entry point.
- Migrate the cross-domain tiebreaker, same-day dedup (Jaccard), and MOC-promotion logic out of `save-conversation` so it isn't re-implemented per caller.
- Preserve current behavior: INDEX linking, daily `## Session Links`, `.INDEX.state` sync, `strict_domains` allow-list validation, `capture_needs_confirmation`.

### References
- `docs/superpowers/specs/2026-06-29-keeper-sole-writer-design.md`
- `docs/superpowers/specs/2026-06-29-keeper-cli-design.md`
- `docs/superpowers/specs/2026-06-29-keeper-insert-design.md`

Not started — tracking only.

nhangen:
Shipped in v1.16.0 (merged to master, `71f1a28`). Two shared libs — `scripts/lib/dedup-scan.sh` (single `tokenize_slug` + Jaccard + same-day scan) and `scripts/lib/allowlist-validate.sh` (taxonomy parse + target validation) — now back save-conversation, create-note, and the vault-librarian; routing prose single-sourced into `obsidian.local.md.example`'s `## Routing Rules` referencing `allowlist_list` (no hardcoded folders). 6 new test suites; full suite green. Follow-ups tracked separately.

PR #34, opened 2026-07-28
Shell-agnostic lib self-location, honest no-config refusal, one tokenizer

### LINKED ISSUE #27: Follow-ups from #26: third tokenizer in vault-scan.sh + allowlist no-config diagnostic
Two non-blocking items surfaced during the #26 final review (both deferred as out-of-scope there):

1. **Third inline tokenizer.** `scripts/lib/vault-scan.sh` `scan_clusters` (~lines 53–61) tokenizes slugs with its own inline logic, feeding the background surfacing path (`vaultkeeper-tick.sh`). #26 unified the dedup + MOC tokenizers onto `tokenize_slug` but did not touch this one (separate path). Pre-existing drift risk — consider routing it through the shared `tokenize_slug` too, or documenting why it must differ.

2. **`allowlist_validate` blank closest-match on config-resolution failure.** When no config resolves, `allowlist_validate` refuses the target (fail-closed, correct) but prints `Closest match: ` with a blank value. Low-harm cosmetic; a distinct "no config resolved — run /obsidian:setup" diagnostic would be more actionable.

Tracking only — neither blocks anything.

nhangen:
Closes #27

#27 tracked two deferred follow-ups from #26. Working them surfaced a third, more serious problem underneath the first, and a fourth alongside it. An 8-agent review panel then found a fifth (a test that could not fail) and corrected my own explanation of the third — see "What the panel changed" below.

**Libs could not find their own siblings under zsh.** `allowlist-validate.sh` and `keeper-bootstrap.sh` located neighbouring scripts with `dirname "${BASH_SOURCE[0]}"`. `BASH_SOURCE` is a bashism — zsh leaves it unset, so that expression evaluates to `dirname ""` = `.` and every self-relative lookup silently became **cwd-relative**. zsh is the runtime: the librarian bootstrap sources these libs and `scripts/keeper` runs under it. From any directory other than `scripts/lib`, `allowlist_validate` resolved no config, failed closed, and **refused every write**.

This was not theoretical — it fired while saving a session note, which is why that save validated the allow-list by hand against the taxonomy table instead of trusting the validator. Both libs now capture their directory at source time via `"${BASH_SOURCE[0]:-$0}"`. zsh's `$0` is the sourced file while it is being read, which is exactly why the capture cannot live inside a function — there, `$0` is the function name.

**The refusal message hid which failure had happened** (item 2 of #27, filed as cosmetic). With no config, both scan loops iterate zero times, so the message read `Closest match: ` with nothing after it — a missing config presenting as a bad target. The allow-list is now read once, and a config whose taxonomy yields no rows gets its own message naming `/obsidian:setup`.

**Clustering carried its own inline tokenizer** (item 1 of #27). `scan_clusters` tokenized slugs inline; #26 unified dedup and MOC promotion onto `tokenize_slug` but not this path, so background surfacing could drift from the paths it feeds. It now calls the shared tokenizer, unconditionally.

**The scanners returned the status of their last test.** `scan_clusters` leaked the result of its final `[ cnt -ge threshold ]` comparison, so a vault whose last-scanned folder had no above-threshold token returned 1 *after writing correct output*. All four `scan_*` now return 0 explicitly, and the contract is stated once in the file header.

## What the panel changed

**I had the `set -e` victim wrong.** The original body said `vaultkeeper-tick.sh` "runs `set -euo pipefail` and assigns from it", implying the tick was aborting. It wasn't. The tick's brace group ends with `if [ -n "$CONFLICTS" ]; then …; fi`, which returns 0 either way, so the scan's status never becomes the group's status. Verified both directions: with that `if` last the assignment survives; delete only that line and the same construct exits 1. The fix is still correct — the tick is protected by which line happens to be last, and nothing recorded that — but the shape actually at risk is a bare `X="$(scan_clusters …)"` under `set -e`.

**A test that could not fail.** `tests/allowlist-validate.sh` asserted `Closest match:[[:space:]]*$`, anchored to end of line, against a message whose text continues past the match. It returned zero hits for every input. And the new empty-taxonomy branch had no coverage at all: with no config, `allowlist_list` fails and the `|| return 1` short-circuits before the list is inspected, so nothing in the suite reached it. Both now covered, both mutation-checked.

**Self-location still misdirected on failure.** Fixing the capture left its own failure path reproducing #27: `cd` failing leaves the var empty, and `dirname ""` → `.` → `cd .` *succeeds* onto the caller's cwd. Either way the user was told "no config resolved — run /obsidian:setup" with a good config on disk, and `keeper-bootstrap` blamed `install-watcher.sh` by name when that file was fine. Both libs now test for the expected sibling and name the install.

**`resolve-config.sh` keeps its bare comparison, deliberately.** It is the last `BASH_SOURCE` site in `scripts/lib/`, which reads like an oversight. Making it `${BASH_SOURCE[0]:-$0}` compares `$0` to itself under zsh, fires the CLI branch while merely being sourced, and the `exit` takes out the sourcing shell — `session-save.sh`, the Stop hook. Verified. Took `${BASH_SOURCE[0]:-}` (unchanged semantics, `set -u` clean) and wrote the reason down.

**Prose drift.** The gate now refuses for two reasons; the librarian and both SKILL.md call sites described one. They promised a "closest match" two of three refusals don't print, and told the LLM to "ask the user where to file" on a *config* fault — where a named folder becomes de-facto authorization to write outside the allow-list.

## Behaviour change to be aware of

**Cluster tokens are now case-folded**, so `Alpha-`, `alpha-`, and `ALPHA-` fall in one cluster. Two consequences on the first tick after this lands:

- Previously-distinct mixed-case tokens merge, cross the threshold, and produce **genuinely new** `CLUSTER` rows — the intended improvement.
- Any cluster whose token was previously emitted mixed-case reads as new against the persisted candidates snapshot, so `Pending.md` gets a **one-time batch of duplicate `- [ ] CLUSTER:` lines** next to rows you may already have acted on. `Pending.md` is hand-edited, so this is worth knowing before it happens rather than after. Not normalizing the snapshot here on purpose — that touches `keeper-state.sh`, which is outside this PR.

## Testing Procedure

1. Check out this branch and run `bash tests/run-all.sh` — expect `ALL PASS` (29 suites).
2. Confirm the self-location coverage runs on your host. `bash tests/allowlist-validate.sh` and `bash tests/vault-scan.sh` should print no `SKIP:` line if zsh is installed; if it isn't, they now say so out loud instead of reporting `ok` for coverage that never ran.
3. Mutation-check the headline fix: `sed -i '' '10s/.*/_av_lib_dir="."/' scripts/lib/allowlist-validate.sh`, then `bash tests/allowlist-validate.sh` → expect `FAIL: bash + foreign cwd`. Restore with `git checkout -- scripts/lib/allowlist-validate.sh`.
4. Mutation-check the new refusal coverage: delete the `if [ -z "$list" ]` block from `allowlist_validate`, run `bash tests/allowlist-validate.sh` → expect `FAIL: empty-taxonomy refusal should name the taxonomy`. Restore.
5. Mutation-check the scanner status: delete the `return 0` at `scripts/lib/vault-scan.sh:40`, run `bash tests/vault-scan.sh` → expect `FAIL: scan_frontmatter_gaps returned 1 on a gap-free vault`. Restore.
6. Confirm the rejected variant is genuinely dangerous: change `resolve-config.sh`'s guard to `${BASH_SOURCE[0]:-$0}` and run `zsh -c '. scripts/lib/resolve-config.sh; echo SURVIVED'` → the echo never prints. Restore.
7. Verify the write gate from a foreign cwd under zsh: `cd / && zsh -c ". <repo>/scripts/lib/allowlist-validate.sh; allowlist_validate 'Daily/2026-07-28.md'"` → exit 0, no output.

## Additional Notes

Follow-ups from the panel, filed separately rather than expanded into this PR: the scheduler orphaning on plugin upgrade (label-only installed check against a version-pinned plist path — the keeper silently stops after every update); `tokenize_slug` not folding spaces, which leaves same-day dedup blind to `beta note one` vs `beta-note-two`; `surfacing_pending_transition` appending without deduping against `Pending.md`; unchecked `find` status reporting a partial scan as complete; and the flat test fixture that never exercises the `.vaultkeeper*` exclusions.

### DIFF
--- a/scripts/lib/allowlist-validate.sh
+++ b/scripts/lib/allowlist-validate.sh
@@ -2,12 +2,19 @@
 # allowlist-validate.sh — Project Taxonomy allow-list parsing + target validation.
 # The taxonomy table in the config is the single source of valid top-level folders.
 
+# Captured at source time, not inside a function: `BASH_SOURCE` is a bashism that
+# zsh leaves unset, and zsh's `$0` is the sourced file only while it is being
+# read — inside a function it is the function name. Getting this wrong made
+# every self-relative lookup cwd-relative under zsh (the librarian and keeper
+# runtime), so config resolution failed and validation refused every write (#27).
+_av_lib_dir="$(cd "$(dirname "${BASH_SOURCE[0]:-$0}")" 2>/dev/null && pwd)"
+
 _allowlist_config() {
   local cfg="${1:-}"
   if [ -z "$cfg" ]; then
-    cfg="$(bash "$(dirname "${BASH_SOURCE[0]}")/resolve-config.sh" 2>/dev/null || true)"
+    cfg="$(bash "${_av_lib_dir}/resolve-config.sh" 2>/dev/null || true)"
   fi
-  [ -n "$cfg" ] && [ -f "$cfg" ] || { printf 'allowlist: no config resolved\n' >&2; return 1; }
+  [ -n "$cfg" ] && [ -f "$cfg" ] || { printf 'allowlist: no config resolved — run /obsidian:setup\n' >&2; return 1; }
   printf '%s\n' "$cfg"
 }
 
@@ -25,7 +32,17 @@ allowlist_list() {
 }
 
 allowlist_validate() {
-  local target="$1" cfg="${2:-}" entry ntarget nentry best="" bestd=9999 d top
+  local target="$1" cfg="${2:-}" entry ntarget nentry best="" bestd=9999 d top list
+  # Read the allow-list once. An empty list means the taxonomy could not be read
+  # at all — a different failure from "the target is not on the list", and
+  # reporting it as the latter (with a blank closest match) sent readers looking
+  # for a typo in a target that was fine (#27).
+  list="$(allowlist_list "$cfg")" || return 1
+  if [ -z "$list" ]; then
+    printf 'Refusing to write to %s — the config has no readable ## Project Taxonomy allow-list. Run /obsidian:setup or add the table.\n' "$target" >&2
+    return 1
+  fi
+
   ntarget="$(printf '%s' "$target" | tr 'A-Z' 'a-z')"; ntarget="${ntarget%/}"
   while IFS= read -r entry; do
     [ -z "$entry" ] && continue
@@ -33,14 +50,18 @@ allowlist_validate() {
     case "$ntarget/" in
       "$nentry"/*) return 0 ;;
     esac
-  done < <(allowlist_list "$cfg")
+  done <<EOF
+$list
+EOF
 
   top="${target%%/*}"
   while IFS= read -r entry; do
     [ -z "$entry" ] && continue
     d="$(_lev "$(printf '%s' "$top" | tr 'A-Z' 'a-z')" "$(printf '%s' "${entry%%/*}" | tr 'A-Z' 'a-z')")"
     if [ "$d" -lt "$bestd" ]; then bestd="$d"; best="$entry"; fi
-  done < <(allowlist_list "$cfg")
+  done <<EOF
+$list
+EOF
 
   printf 'Refusing to write to %s — top-level folder is not in the allow-list. Closest match: %s. Add it to ## Project Taxonomy or correct the target.\n' "$target" "$best" >&2
   return 1
--- a/scripts/lib/allowlist-validate.sh
+++ b/scripts/lib/allowlist-validate.sh
@@ -43,10 +43,12 @@ allowlist_list() {
 
 allowlist_validate() {
   local target="$1" cfg="${2:-}" entry ntarget nentry best="" bestd=9999 d top list
-  # Read the allow-list once. An empty list means the taxonomy could not be read
-  # at all — a different failure from "the target is not on the list", and
-  # reporting it as the latter (with a blank closest match) sent readers looking
-  # for a typo in a target that was fine (#27).
+  # An empty list means no valid taxonomy rows parsed — a missing heading, or a
+  # table with only a header. Distinct from "the target is not on the list", which
+  # offers a closest match; reporting the former as the latter points the reader at
+  # a typo in a target that was fine. A config that is present but unreadable does
+  # not arrive here: awk fails, allowlist_list propagates, and the `|| return 1`
+  # below short-circuits.
   list="$(allowlist_list "$cfg")" || return 1
   if [ -z "$list" ]; then
     printf 'Refusing to write to %s — the config has no readable ## Project Taxonomy allow-list. Run /obsidian:setup or add the table.\n' "$target" >&2
--- a/scripts/lib/allowlist-validate.sh
+++ b/scripts/lib/allowlist-validate.sh
@@ -12,6 +12,16 @@ _av_lib_dir="$(cd "$(dirname "${BASH_SOURCE[0]:-$0}")" 2>/dev/null && pwd)"
 _allowlist_config() {
   local cfg="${1:-}"
   if [ -z "$cfg" ]; then
+    # Check the capture landed somewhere real before trusting it. Testing for the
+    # sibling rather than for an empty string catches both failure shapes: `cd`
+    # failing (empty) and `dirname ""` -> "." -> `cd .` succeeding onto the
+    # caller's cwd. Without this, either one resolves no config and reports it as
+    # a missing config, sending the reader to /obsidian:setup for what is a broken
+    # install — the same misdirection this file's own refusal was fixed for.
+    if [ ! -f "${_av_lib_dir}/resolve-config.sh" ]; then
+      printf 'allowlist: cannot find resolve-config.sh beside this lib (looked in "%s") — the install is broken, not the config\n' "$_av_lib_dir" >&2
+      return 1
+    fi
     cfg="$(bash "${_av_lib_dir}/resolve-config.sh" 2>/dev/null || true)"
   fi
   [ -n "$cfg" ] && [ -f "$cfg" ] || { printf 'allowlist: no config resolved — run /obsidian:setup\n' >&2; return 1; }
--- a/scripts/lib/keeper-bootstrap.sh
+++ b/scripts/lib/keeper-bootstrap.sh
@@ -13,7 +13,11 @@
 # crontab line, so each host installs its own scheduler the first time normal
 # workflow runs on it. Opt out with `keeper_autostart: false` in the config.
 
-_kb_scripts() { cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd; }
+# Captured at source time — see the note in allowlist-validate.sh: zsh has no
+# BASH_SOURCE, and its `$0` only names the sourced file while it is being read.
+_kb_lib_dir="$(cd "$(dirname "${BASH_SOURCE[0]:-$0}")" 2>/dev/null && pwd)"
+
+_kb_scripts() { printf '%s\n' "${_kb_lib_dir%/*}"; }
 
 # Single source of truth for the namespace label: install-watcher.sh owns it.
 keeper_label() { bash "$(_kb_scripts)/install-watcher.sh" label 2>/dev/null; }
--- a/scripts/lib/keeper-bootstrap.sh
+++ b/scripts/lib/keeper-bootstrap.sh
@@ -17,7 +17,16 @@
 # BASH_SOURCE, and its `$0` only names the sourced file while it is being read.
 _kb_lib_dir="$(cd "$(dirname "${BASH_SOURCE[0]:-$0}")" 2>/dev/null && pwd)"
 
-_kb_scripts() { printf '%s\n' "${_kb_lib_dir%/*}"; }
+# Returns non-zero rather than printing an empty path. The old `cd ../ && pwd`
+# propagated its failure; a bare parameter strip cannot, and both callers use the
+# result unchecked to build `bash "$scripts/..."` command lines. Testing for the
+# sibling covers a `cd` that failed (empty) and one that succeeded onto the
+# caller's cwd; it also covers `${x%/*}` having no root case, where "/" strips to "".
+_kb_scripts() {
+  local d="${_kb_lib_dir%/*}"
+  [ -n "$d" ] && [ -f "$d/install-watcher.sh" ] || return 1
+  printf '%s\n' "$d"
+}
 
 # Single source of truth for the namespace label: install-watcher.sh owns it.
 keeper_label() { bash "$(_kb_scripts)/install-watcher.sh" label 2>/dev/null; }
@@ -60,6 +69,15 @@ keeper_ensure_active() {
   autostart="$(grep '^keeper_autostart:' "$cfg" | head -1 | sed 's/^keeper_autostart:[[:space:]]*//')"
   [ "$autostart" = "false" ] && return 0
 
+  # Resolve the scripts dir before the label, so a lib that cannot find its own
+  # siblings says that, instead of surfacing as "install-watcher.sh broken?" and
+  # naming a file that is perfectly fine.
+  local scripts
+  if ! scripts="$(_kb_scripts)"; then
+    printf 'vaultkeeper: cannot find the scripts dir beside this lib (looked beside "%s"); activation skipped\n' "$_kb_lib_dir" >&2
+    return 0
+  fi
+
   # A missing/empty label means the installer is broken — that's an error to
   # report, not a "not installed" signal to barrel past into a doomed install.
   local label; label="$(keeper_label)"
@@ -70,7 +88,6 @@ keeper_ensure_active() {
 
   keeper_scheduler_installed && return 0
 
-  local scripts; scripts="$(_kb_scripts)"
   local tick="$scripts/vaultkeeper-tick.sh"
   if ! bash "$scripts/seed-frontmatter-schema.sh" "$vault" "$cfg" >/dev/null 2>&1; then
     printf 'vaultkeeper: frontmatter-schema seed failed; using default (tags type)\n' >&2
--- a/scripts/lib/vault-scan.sh
+++ b/scripts/lib/vault-scan.sh
@@ -18,10 +18,15 @@
 # Directory captured at source time; see the note in allowlist-validate.sh for
 # why BASH_SOURCE alone is not enough (zsh).
 _vs_lib_dir="$(cd "$(dirname "${BASH_SOURCE[0]:-$0}")" 2>/dev/null && pwd)"
-. "${_vs_lib_dir}/dedup-scan.sh" 2>/dev/null || {
-  printf 'vault-scan: cannot source dedup-scan.sh from "%s" — clustering has no tokenizer\n' "$_vs_lib_dir" >&2
+# Pre-test, not `. file || handler`: `.` is a POSIX special builtin, so under a
+# `set -e` caller — which vaultkeeper-tick.sh is, and it is the only production
+# sourcer — the shell exits at the failed source and the handler never runs. Same
+# shape as the sibling check in allowlist-validate.sh.
+if [ ! -r "${_vs_lib_dir}/dedup-scan.sh" ]; then
+  printf 'vault-scan: cannot read dedup-scan.sh beside this lib (looked in "%s") — clustering has no tokenizer\n' "$_vs_lib_dir" >&2
   return 1 2>/dev/null || exit 1
-}
+fi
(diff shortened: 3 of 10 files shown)


tests/allowlist-validate.sh
```diff
@@ -36,5 +36,28 @@ if allowlist_validate "Dailyish/x.md" "$CFG" 2>/dev/null; then fail "Dailyish mu
...
+  # silently became cwd-dependent. Under zsh from any other cwd, config
+  # resolution failed and strict validation refused every target.
+  OBSIDIAN_LOCAL_MD="$CFG" zsh -c "cd / && . '$ROOT_DIR/scripts/lib/allowlist-validate.sh'; allowlist_validate 'Daily/2026-06-29.md'" 2>"$TMP/zerr2" \
+    || { cat "$TMP/zerr2" >&2; fail "zsh + foreign cwd: lib could not locate its own resolve-config.sh"; }
+  # Same shape under bash, for symmetry — a cwd-independent lib must not care.
+  OBSIDIAN_LOCAL_MD="$CFG" bash -c "cd / && . '$ROOT_DIR/scripts/lib/allowlist-validate.sh'; allowlist_validate 'Daily/2026-06-29.md'" 2>"$TMP/berr" \
+    || { cat "$TMP/berr" >&2; fail "bash + foreign cwd: lib could not locate its own resolve-config.sh"; }
+fi
+
+# No config anywhere: the refusal must name the actual problem. Reporting a
```
nhangen, line 58: This assertion can't fire. It greps `Closest match:[[:space:]]*$` but production prints `... Closest match: %s. Add it to ## Project Taxonomy ...`, so there's always text after the match. Verified 0 hits against the real refusal string.

And the block never reaches the new branch anyway — with no config, `allowlist_list` fails, so `|| return 1` on line 40 returns before `[ -z "$list" ]`. Replacing the new message with a nonsense string leaves the suite green.

What's needed: two fixtures, one config with no `## Project Taxonomy` section and one with a header + separator and zero data rows, both asserting `no readable ## Project Taxonomy allow-list` and a non-zero exit. Then either repair the anchored grep or drop it.

tests/allowlist-validate.sh
```diff
@@ -36,5 +36,28 @@ if allowlist_validate "Dailyish/x.md" "$CFG" 2>/dev/null; then fail "Dailyish mu
 
 if command -v zsh >/dev/null 2>&1; then
   zsh -c ". '$ROOT_DIR/scripts/lib/allowlist-validate.sh'; allowlist_list '$CFG'" >/dev/null 2>"$TMP/zerr" || { cat "$TMP/zerr" >&2; fail "allowlist-validate broke under zsh"; }
+
+  # The arm above passes $CFG explicitly, so it never exercises self-resolution —
+  # which is how the real failure hid. The lib locates resolve-config.sh via
```
nhangen: This is the only thing pinning either self-location fix, and it disappears silently on a host without zsh — `run-all.sh` just prints `ok`. Worse, the bash symmetry arm below is nested inside this `if`, so on a zsh-less host neither arm runs, even though the bash one does catch `_av_lib_dir="."`.

Hoist the bash arm out, and print a visible SKIP in the `else` so absent coverage looks absent.

Also the comment reads present-tense ("the lib locates resolve-config.sh via BASH_SOURCE") in the file that stops it doing that — should be "used to locate".

tests/vault-scan.sh
```diff
@@ -49,6 +49,42 @@ has "CLUSTER"$'\t'"Projects"$'\t'"weekly"$'\t'"3" "$CLUSTERS"
...
+# is case folding: these three belong to one cluster, and an uppercased token is
+# not its own separate one.
+mkdir -p "$V/Casing"
+: > "$V/Casing/Alpha-first.md"
+: > "$V/Casing/alpha-second.md"
+: > "$V/Casing/ALPHA-third.md"
+CCLUST="$(scan_clusters "$V" 3)"
+has "CLUSTER"$'\t'"Casing"$'\t'"alpha"$'\t'"3" "$CCLUST"
+grep -qF 'ALPHA' <<<"$CCLUST" && fail "cluster tokens must be case-folded, got:"$'\n'"$CCLUST"
+# scan_* report through stdout, so their exit status must be success even when
```
nhangen: The last clause here is wrong. `vaultkeeper-tick.sh` does run `set -euo pipefail` and does assign from these scans, but it was never vulnerable: the brace group ends with `if [ -n "$CONFLICTS" ]; then ...; fi`, which always returns 0, so the group's status is never `scan_clusters`'.

Reproduced the exact construct — with that trailing `if`, rc=0; delete only that line, rc=1.

The real exposure is a bare `X="$(scan_clusters …)"` under `set -e`, which is the shape the plan doc uses. Worth saying that instead, and worth noting the tick is protected only by line ordering.

tests/vault-scan.sh
```diff
@@ -49,6 +49,42 @@ has "CLUSTER"$'\t'"Projects"$'\t'"weekly"$'\t'"3" "$CLUSTERS"
...
+# they find nothing. scan_clusters used to return the status of its last
+# threshold test: one trailing folder with no above-threshold token made it
+# return 1, which aborts any `set -e` caller — and vaultkeeper-tick.sh runs
+# `set -euo pipefail` and assigns from it.
+mkdir -p "$V/ZZlast"
+: > "$V/ZZlast/solitary-note.md"
+CRC=0; ( set -e; scan_clusters "$V" 3 >/dev/null ) || CRC=$?
+[ "$CRC" = "0" ] || fail "scan_clusters returned $CRC with a below-threshold trailing folder"
+GRC=0; ( set -e; scan_frontmatter_gaps "$V" "tags type" >/dev/null ) || GRC=$?
```
nhangen, line 76: `GRC` can't fail — the fixture always contains a gap file, so the loop's last body command succeeds regardless. Deleting the `return 0` on `scan_frontmatter_gaps` leaves the whole suite green.

The guarded failure is real, though: on a vault where every file is complete, the body ends `[ -n "$miss" ] && printf`, so the `while` returns 1. Needs a second all-complete fixture to become coverage.

`URC` and `ARC` are a different problem — those two functions can't return non-zero at all, so those assertions can never fail by construction.

scripts/lib/vault-scan.sh
```diff
@@ -26,6 +32,7 @@ scan_unfiled() {
   while IFS= read -r f; do
     printf 'UNFILED\t%s\n' "${f#"$vault"/}"
   done < <(find "$vault/Inbox" -type f -name '*.md')
+  return 0   # reports on stdout; a failed final test must not become the status
```
nhangen: This comment is true on `scan_frontmatter_gaps` and `scan_clusters` and false here and on `scan_open_asks`. `scan_unfiled`'s loop body is a bare `printf` — there's no test to fail; `scan_open_asks`' body is an `if`, which exits 0 either way. Neither ever returned non-zero.

Keeping `return 0` on all four is fine as a uniform contract, but four verbatim copies of a claim that's wrong on two of them will mislead. One line in the header stating the contract would do it.

scripts/lib/vault-scan.sh
```diff
@@ -2,7 +2,12 @@
 # vault-scan.sh — compute surfacing candidates from the vault. .md only;
 # excludes .obsidian/, .trash/, .git/, .vaultkeeper-quarantine/. Never reads
 # .base files (they are not .md, so the find filters exclude them).
-# Requires frontmatter.sh.
+# Requires frontmatter.sh. Sources dedup-scan.sh for the shared tokenize_slug so
+# clustering cannot drift from dedup/MOC — vaultkeeper-tick.sh does not source it.
+# Directory captured at source time; see the note in allowlist-validate.sh for
+# why BASH_SOURCE alone is not enough (zsh).
+_vs_lib_dir="$(cd "$(dirname "${BASH_SOURCE[0]:-$0}")" 2>/dev/null && pwd)"
+command -v tokenize_slug >/dev/null 2>&1 || . "${_vs_lib_dir}/dedup-scan.sh"
```
nhangen: The header two lines up says clustering "cannot drift" from dedup/MOC, and this guard is exactly the door it drifts through — any caller with a `tokenize_slug` already in scope keeps its own and the canonical one never loads. A caller defining `tokenize_slug(){ printf BOGUS; }` yields `CLUSTER D BOGUS 3`. `command -v` also matches PATH executables and aliases, not just functions.

`dedup-scan.sh` is three pure function definitions with no source-time state, so sourcing unconditionally is idempotent and costs one file read. Either drop the guard or soften the comment to "uses the shared tokenizer".

Separately: this is the first lib in the tree to source its own dependency — every other consumer sources its libs explicitly (`vaultkeeper-tick.sh` lists seven).

scripts/lib/allowlist-validate.sh
```diff
@@ -2,12 +2,19 @@
 # allowlist-validate.sh — Project Taxonomy allow-list parsing + target validation.
 # The taxonomy table in the config is the single source of valid top-level folders.
 
+# Captured at source time, not inside a function: `BASH_SOURCE` is a bashism that
+# zsh leaves unset, and zsh's `$0` is the sourced file only while it is being
+# read — inside a function it is the function name. Getting this wrong made
+# every self-relative lookup cwd-relative under zsh (the librarian and keeper
+# runtime), so config resolution failed and validation refused every write (#27).
+_av_lib_dir="$(cd "$(dirname "${BASH_SOURCE[0]:-$0}")" 2>/dev/null && pwd)"
```
nhangen, line 10: The `2>/dev/null` here swallows the one signal that says self-location failed, and the failure then presents as the wrong cause — the exact class of misdiagnosis item 2 of #27 was about. Two shapes:

- `dirname` yields `""` → `cd ""` is a no-op and `pwd` returns the **caller's cwd**, i.e. the #27 behaviour, silently.
- directory gone or unreadable → `cd` fails, the var is empty, and with a perfectly good config on disk the user is told `no config resolved — run /obsidian:setup`. Setup can't fix a lib that lost its path.

One non-empty check per lib turns it into a named failure. Same applies to `_kb_lib_dir` and `_vs_lib_dir`.

scripts/lib/allowlist-validate.sh
```diff
@@ -25,22 +32,36 @@ allowlist_list() {
 }
 
 allowlist_validate() {
-  local target="$1" cfg="${2:-}" entry ntarget nentry best="" bestd=9999 d top
+  local target="$1" cfg="${2:-}" entry ntarget nentry best="" bestd=9999 d top list
+  # Read the allow-list once. An empty list means the taxonomy could not be read
```
nhangen: "could not be read at all" isn't what an empty list means — it means no valid rows parsed, which covers a missing heading *and* a present-but-empty table. The genuinely unreadable config (`chmod 000`) never gets here: `awk` fails, `allowlist_list` propagates non-zero, and line 40 returns first, so what actually surfaces is a raw `awk: can't open file` with no refusal line at all.

The code's own wording ("no readable … allow-list") is more accurate than the comment. Also `# Read the allow-list once.` just narrates the assignment below it.

scripts/lib/vault-scan.sh
```diff
@@ -52,13 +60,14 @@ scan_clusters() {
...
-          printf '%s' "$tok" | grep -qE '^[0-9]+$' && continue
-          [ "${#tok}" -le 1 ] && continue
-          printf '%s\n' "$tok"
-        done
+        # Shared tokenizer, so clustering agrees with dedup and MOC promotion
+        # instead of drifting as a third inline copy (#27). Spaces are folded to
+        # the same separator first — tokenize_slug splits on `-` only, and this
+        # vault has filenames with spaces. `sort -u` keeps the count "distinct
+        # files containing the token", not total occurrences.
+        printf '%s\n' "$base" | tr ' ' '-' | while IFS= read -r slug; do
```
nhangen, line 86: Case-folding is the right change, but it's a migration and the PR doesn't say so. The candidates snapshot on disk holds the previous scan's raw `CLUSTER\t<dir>\t<tok>\t<cnt>` lines, and `surfacing_pending_transition` diffs against it, so on the first tick after this lands every previously mixed-case token reads as new and gets appended to `Pending.md` next to the row you may already have acted on. Merged tokens also cross the threshold and produce genuinely new clusters (confirmed on a prod-shaped fixture: a new `quantum` row that master didn't emit).

The new clusters are the point. The duplicates aren't. Either normalize the snapshot once on read or say in the body that the first tick appends a one-time batch — `Pending.md` is hand-edited.

Also: "a third inline copy" miscounts. On master there were exactly two implementations, this one and `dedup-scan.sh`'s.

scripts/lib/keeper-bootstrap.sh
```diff
@@ -13,7 +13,11 @@
 # crontab line, so each host installs its own scheduler the first time normal
 # workflow runs on it. Opt out with `keeper_autostart: false` in the config.
 
-_kb_scripts() { cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd; }
+# Captured at source time — see the note in allowlist-validate.sh: zsh has no
+# BASH_SOURCE, and its `$0` only names the sourced file while it is being read.
+_kb_lib_dir="$(cd "$(dirname "${BASH_SOURCE[0]:-$0}")" 2>/dev/null && pwd)"
```
nhangen, line 18: Two things on this pair.

The capture's swallowed `cd` (same as in allowlist-validate) leaves `_kb_lib_dir` empty, and then `_kb_scripts` prints an empty string and **returns 0** — the old `cd … && pwd` propagated the failure. Both callers use it unchecked, so `keeper_label` runs `bash "/install-watcher.sh"`, and the diagnostic that surfaces blames `install-watcher.sh` when the installer is fine. Net effect is the keeper never activating on that host with a message pointing at the wrong file.

Also `${_kb_lib_dir%/*}` isn't `cd ..` — it has no root case, so `/` strips to empty rather than `/`. Only reachable once the capture already failed, so a non-empty guard on the capture covers both.

nhangen (commented):
Eight-agent panel on this. The three diagnoses hold up and the fixes are at the right level, but the panel found one thing that has to change before merge and one claim in my own PR body that's flat wrong.

**The `set -euo pipefail` justification is false.** I wrote that `vaultkeeper-tick.sh` "runs `set -euo pipefail` and assigns from it," implying the tick was aborting. It wasn't. The brace group's last command is `if [ -n "$CONFLICTS" ]`, which always returns 0, so `scan_clusters`' status never becomes the group's status. Reproduced both ways: with the trailing `if`, the assignment survives; delete that one line and the same construct exits 1. So the tick is protected purely by the accident of which line happens to be last — that's the honest reason to fix this, and it's a better one, but the body needs rewriting.

**The blocker is a test that cannot fail.** `tests/allowlist-validate.sh` greps `Closest match:[[:space:]]*$` — anchored to end of line — against a message whose text continues after `%s`. It returns 0 matches for any input. And in the no-config case, `allowlist_list` fails so line 40's `|| return 1` short-circuits before the new `[ -z "$list" ]` branch ever runs. `grep -rn "no readable" tests/` finds nothing. The fix itself is correct and fails closed — I drove the branch directly — but nothing in the suite goes there, and the comment above it says otherwise. In a repo with no CI, that suite is the whole gate.

Two more that need a note rather than an inline anchor:

- `tests/vault-scan.sh` has no zsh or foreign-cwd arm for `vault-scan.sh`'s own capture. Reverting it leaves the suite green — it's the one of the three whose failure is worst (no `tokenize_slug`, and the tick dies at source time) and the only one with no failing mutant.
- `scripts/lib/resolve-config.sh:48` is the last un-swept `BASH_SOURCE` in `scripts/lib/`. Leaving the bare form is the right call and I stand by it, but it needs a comment saying why, or the next tidy-up reintroduces a shell-killer. Use `${BASH_SOURCE[0]:-}` if changing it — **not** `${BASH_SOURCE[0]:-$0}`, which makes the CLI branch fire under zsh sourcing and `exit` the caller. `session-save.sh` sources that file.
- `agents/vault-librarian.md:71` and both SKILL.md call sites still describe one refusal shape. Two of the three now print no closest match, and the librarian's blanket "ask the user where to file" is the wrong response to a config fault — it turns a missing config into a folder question a user can answer, which is how an unvalidated write gets authorized.

Follow-ups worth their own issues, not this PR: the scheduler orphaning on plugin upgrade (label-only installed check vs. a version-pinned plist path — the keeper silently stops after every update); `tokenize_slug` not folding spaces, which leaves same-day dedup blind to `beta note one` vs `beta-note-two`; `surfacing_pending_transition` appending without deduping against `Pending.md`; and unchecked `find` status reporting a partial scan as complete.

(Verdict is COMMENT rather than REQUEST_CHANGES only because GitHub won't let me request changes on my own PR.)

Issue #35, opened 2026-07-28
Plugin upgrade orphans the keeper scheduler


nhangen:
Found by the PR #34 review panel (domain lens). Pre-existing, not introduced by #34.

`keeper_scheduler_installed` (`scripts/lib/keeper-bootstrap.sh`) detects an installed scheduler by matching only the **label** in `launchctl list` / `crontab -l`. But `install-watcher.sh` bakes the **absolute, version-pinned** tick path into the plist / cron line:

    ~/.claude/plugins/cache/nhangen/obsidian/<version>/scripts/vaultkeeper-tick.sh

A plugin update replaces that directory wholesale. `launchctl` still lists the label, so `keeper_ensure_active` early-returns forever while the program it points at no longer exists. Net effect: **the keeper silently stops ticking after every plugin upgrade, on every host, and self-activation never repairs it.**

`resolve-config.sh:4-9` documents exactly this hazard for the *config*. Nothing does it for the scheduler.

Fix: compare the installed program path (plist `ProgramArguments`, or the cron line) against the current `$tick` and reinstall on mismatch. The plist is already rendered from one template, so this is cheap.

Detection surface today is `staleness_banner` in `/obsidian:ask`, which reports but repairs nothing.

nhangen:
Re-read this after learning that Claude Code exposes `${CLAUDE_PLUGIN_ROOT}` (always the *active* version) and `${CLAUDE_PLUGIN_DATA}` (survives updates), documented in `plugins-reference`. First instinct was that this issue collapses into "the plist should have used the substitution." **It does not** — worth writing down so nobody tries it.

The substitution is performed by Claude Code where Claude Code reads config: hook `command`/`args`/`env`, MCP and LSP server config, and skill/agent markdown. `launchctl` and `cron` are outside that layer. `ProgramArguments` is a literal argv — launchd performs no variable expansion — so a plist containing `${CLAUDE_PLUGIN_ROOT}/scripts/vaultkeeper-tick.sh` would try to exec a directory named `${CLAUDE_PLUGIN_ROOT}`. Same for a crontab line, which gets `/bin/sh` but not our env.

`install-watcher.sh:21` renders `<string>${tick}</string>`, a shell variable expanded **at install time** to an absolute path. That is the crux: whatever `$tick` resolves to gets frozen into the plist, and if it points inside the versioned cache dir it dies on the next update.

[...]

Issue #4, opened 2026-05-09
Cross-domain routing tiebreaker (Career/Finance/etc.)


nhangen:
## Problem

When a topic matches multiple plausible routes, the skill picks one without surfacing siblings. The 2026-05-09 vault reorg found three parallel trees for career-related notes:

- `Career/`
- `Awesome Motive/career/`
- `Personal/Job Search/`

All three were active simultaneously because each save resolved differently depending on the topic-hint and the recent-domain heuristic. The user had to manually consolidate.

Same shape, smaller scale: investing/finance notes scattered across `Personal/`, `Plans/`, and `Projects/Personal/` until a `Personal/Finance/` route was added.

## Proposed change

1. When more than one route matches above a confidence threshold, list the candidates and the most-specific top-level domain wins.
2. Before writing, scan sibling candidates (e.g. `Career/`, `Awesome Motive/career/`) for existing notes with overlapping slugs and warn if found.
3. Document a precedence rule in `obsidian.local.md` — current rule has no tiebreaker.

## Acceptance

- Writing a "career strategy" note when both `Career/` and `Awesome Motive/career/` could match prompts the user once and remembers the choice for the session.
- The `obsidian.local.md` schema gains a `precedence:` field per route.

Related to #ALIAS-FOLDER-VALIDATION (issue 3).

Surfaced by: vault reorg 2026-05-09.

Issue #3, opened 2026-05-09
Same-day dedup check before creating new session note


nhangen:
## Problem

The `obsidian:save` skill writes a new file even when a same-day note for the same topic/domain already exists. Recent example from the 2026-05-09 vault reorg:

- `Personal/Reflections/2026-04-22-candid-vault-self-assessment.md`
- `Personal/Reflections/2026-04-22-vault-self-analysis.md`

Two passes on the same reflection on the same day, written as separate files. Cleanup required a manual merge.

## Proposed change

Before writing a new note:

1. Glob the target directory for `<YYYY-MM-DD>-*.md`.
2. If matches exist, score topic similarity against the proposed slug (slug overlap or simple keyword set intersection is enough — no embedding needed for v1).
3. On a strong match, offer two paths: append a new `## <HH:MM> — <topic>` section to the existing file, or proceed with a new file (with a one-line confirmation showing the existing file path).

## Acceptance

- Writing a second same-day note in the same folder triggers the dedup prompt.
- Append mode preserves frontmatter and adds a timestamped section.
- New-file mode is still available with a single confirmation.

Surfaced by: vault reorg 2026-05-09 (`Reference/vault-reorg-2026-05-09.md` in nhangen's vault).

PR #41, opened 2026-07-28
keeper-bootstrap: report config-write, temp-file, and installer failures

### LINKED ISSUE #40: keeper-bootstrap swallows its config writes and its installer diagnostic
From the PR #34 review panel. Pre-existing; #34 fixed the self-location half of this file and left these.

**1. Config writes are silent on failure.** `scripts/lib/keeper-bootstrap.sh`:

    _kb_cfg_ensure keeper_autostart true "$cfg" || true
    _kb_cfg_ensure keeper_interval_secs "$interval" "$cfg" || true

`_kb_cfg_ensure` returns 1 on `mktemp` failure and propagates `awk`/`mv` failure. All of it is discarded. The contrast is two lines above: the frontmatter-schema seed reports its failure *and* names the fallback it is using. So a config that silently never gains `keeper_interval_secs` just runs at the 900s default forever with no trace. Keep the non-fatal semantics — a Stop hook must not abort — but print something.

**2. `_kb_cfg_ensure` leaks its temp file** on the path where `awk` succeeds and `mv` fails. Needs `rm -f "$tmp"`.

**3. `keeper_label` discards the only diagnostic that distinguishes two causes:**

    keeper_label() { bash "$(_kb_scripts)/install-watcher.sh" label 2>/dev/null; }

Missing, non-executable, and genuinely broken all present as an empty string. #34 fixed the specific case where the *lib* could not find the installer (that now reports separately), so what remains is real installer breakage — capture its stderr and include it in the refusal instead of guessing "install-watcher.sh broken?".

nhangen:
Closes #40

Three silent-failure fixes in `scripts/lib/keeper-bootstrap.sh`, one commit each, plus one found while verifying them.

**1. A failed config write names the key** (`fa724c9`). `_kb_cfg_ensure ... || true` discarded every failure. Both failure modes do print something of their own — `mktemp` reports `mkstemp failed on ...` and `mv` reports "Permission denied", and line 64 redirects neither — but nothing said *which key was lost or what happens next* — so a config that never gains `keeper_interval_secs` runs at the compiled-in 900s default forever with no trace. Both keys are now reported by name, matching the schema-seed line above them which already names its fallback. `|| printf` still evaluates to 0, so the Stop hook stays unabortable.

**2. No temp-file leak on a failed swap** (`3c69d7f`). When `awk` succeeded and `mv` failed, the fully-rendered config stayed in `TMPDIR`. Reachable once per host, not once per session: `keeper_scheduler_installed && return 0` sits above this, so every session after the first returns before reaching it. The leak is one rendered config, on first activation, on a host whose config dir is unwritable. The cleanup clause also covers the `awk`-failure path, which leaked a partial render.

**3. A broken installer is quoted, not guessed at** (`c2b1a93`). `keeper_label` swallowed `install-watcher.sh`'s stderr, so a syntax error, a missing dependency, and an unreadable file all arrived as an empty string and the refusal guessed `install-watcher.sh broken?`. #34 split out the case where the *lib* can't find the installer; what remained was real installer breakage, and its stderr is the only thing identifying which. `keeper_label` now lets that through, and fails when `_kb_scripts` does instead of interpolating an unchecked path into `bash "$(_kb_scripts)/..."` and calling `/install-watcher.sh`.

**4. The refusals can actually be reached under errexit** (`49ed251`) — not in the issue; found while verifying #3. Under `set -euo pipefail` the broken-installer path aborted the caller and printed **nothing**, making it strictly worse than the guess it replaced. Two assignments took their status from a command that legitimately fails: `label="$(keeper_label ...)"` (broken installer exits non-zero) and `autostart="$(grep '^keeper_autostart:' ...)"` (a config without the key makes grep exit 1, and `pipefail` promotes it — that's the ordinary first-run config). Both now `|| <var>=""`. That is almost purely a reachability change, with one edge: discarding stdout on a non-zero exit means an installer that prints a label *and* exits non-zero now refuses where it previously proceeded. Today's `install-watcher.sh` cannot do that (`label)` is a bare `printf`), so nothing in tree behaves differently. Same failure class as the `.`-under-errexit finding on #34: an error handler the shell never reaches.

`session-save.sh` is `set -uo pipefail` and calls this with `|| true`, so nothing in production was aborting today — item 4 is about the next sourcer.

**5. The errexit guard at the third call site** (`19fbf32`) — from the review panel on this PR. `keeper_scheduler_installed`'s own `label="$(keeper_label)"` was the call #4 missed, so a direct call under `set -euo pipefail` died on that line and never reached the `[ -n "$label" ] || return 1` below it: `rc=3` against a broken installer, `rc=1` with nothing printed against an orphaned lib. Production escaped it only because the call site is the left operand of `&&`, which suppresses errexit positionally. It also drops the installer's stderr — a predicate has no channel to explain a fault, so that text was arriving unprefixed at whoever asked a yes/no question.

**6. The installer is quoted only when it failed** (`1e6b4a8`). Capturing its stderr unconditionally put a `mktemp`/`tr`/`rm` and a *second* `install-watcher.sh` process on the session-end path of every activated host, to quote an installer that was working. The temp file now appears only on the branch that is already refusing to act, which deletes the untested mktemp-failure `else` arm instead of testing it, and the predicate takes the resolved label as an argument so it stops re-asking. The captured text is bounded (`tr -c '[:print:]'` + `head -c 300`) — it is third-party output headed for a hook's stderr, and 200 KB arrived as one live-ANSI line.

**7. Four comments corrected** (`a59f870`). Two were falsified by commits 1-4 above ("both callers use the result unchecked"; the quoted `install-watcher.sh broken?` refusal), one named the wrong actor ("the sourcing hook" — `session-save.sh` isn't errexit and calls with `|| true`), one asserted the per-session frequency corrected in item 2. Plus a `chmod` before the test teardown's `rm -rf`.

Panel also found one thing outside this diff worth its own issue: on this machine 710 of 710 keeper ticks log `vault-scan.sh: ... Too many open files` and then report `vaultkeeper: scan complete`. Filed separately, along with the temp-swap invariant still unfixed in three sibling libs, the plist clobber on install, the `|| autostart=""` pipeline scope, and the never-retried schema seed.

## Testing Procedure

1. Check out this branch and run `bash tests/run-all.sh` — expect `ALL PASS` (28 suites).
2. Run `bash tests/keeper-bootstrap.sh` alone. `mv: ... Permission denied` lines on stderr are expected: two arms deliberately make the config unwritable.
3. Confirm each assertion pins its fix — revert one line, watch the named test fail, restore:
   - restore `|| true` on the two `_kb_cfg_ensure` calls → `failed keeper_autostart write was silent`
   - drop `|| { rm -f "$tmp"; return 1; }` from `_kb_cfg_ensure` → `_kb_cfg_ensure leaked 1 temp file(s)`
   - restore `2>/dev/null` in `keeper_label` → `installer stderr was discarded`
   - drop either `|| label=""` / `|| autostart=""` → `[set -euo pipefail;] ensure_active must return 0`
4. Verify the errexit arm is real, not theory: `bash -c "set -euo pipefail; . scripts/lib/keeper-bootstrap.sh; ..."` against a stub installer that exits non-zero returned `rc=1` with empty output before `49ed251`.
5. zsh checked separately (the keeper runtime has no `BASH_SOURCE`): label resolves, and the broken-installer refusal prints with the installer's own text.

## Additional Notes / HelpScout Tickets

Pre-existing issues in a file #34 touched; #34 fixed the self-location half and left these. Findings originated in the PR #34 review panel.

While verifying, confirmed this host's launchd agent points at the version-resilient delegator `~/.claude/hooks/obsidian-vaultkeeper-tick.sh` (resolves newest installed version), not a version-pinned path — so #35's orphaning does not apply to this host as currently installed. #35 is still valid for a fresh `install-watcher.sh` install, which does bake an absolute versioned path.

### DIFF
--- a/scripts/lib/keeper-bootstrap.sh
+++ b/scripts/lib/keeper-bootstrap.sh
@@ -92,8 +92,15 @@ keeper_ensure_active() {
   if ! bash "$scripts/seed-frontmatter-schema.sh" "$vault" "$cfg" >/dev/null 2>&1; then
     printf 'vaultkeeper: frontmatter-schema seed failed; using default (tags type)\n' >&2
   fi
-  _kb_cfg_ensure keeper_autostart true "$cfg" || true
-  _kb_cfg_ensure keeper_interval_secs "$interval" "$cfg" || true
+  # Non-fatal, but not silent: `|| true` discarded a failed mktemp/awk/mv, and a
+  # config that never gains keeper_interval_secs then runs at the compiled-in
+  # default forever with nothing to explain why. Name the key and the
+  # consequence, the way the schema seed above names its fallback. `|| printf`
+  # keeps the exit status 0, so the hook still cannot be aborted by this.
+  _kb_cfg_ensure keeper_autostart true "$cfg" \
+    || printf 'vaultkeeper: could not write keeper_autostart to %s; opt-out state is unrecorded\n' "$cfg" >&2
+  _kb_cfg_ensure keeper_interval_secs "$interval" "$cfg" \
+    || printf 'vaultkeeper: could not write keeper_interval_secs to %s; readers will fall back to the %ss default\n' "$cfg" "$interval" >&2
 
   local ok=""
   if [ -n "${VAULTKEEPER_INSTALL:-}" ]; then
--- a/scripts/lib/keeper-bootstrap.sh
+++ b/scripts/lib/keeper-bootstrap.sh
@@ -30,10 +30,11 @@ _kb_scripts() {
 
 # Single source of truth for the namespace label: install-watcher.sh owns it.
 # Deliberately does NOT swallow the installer's stderr — that text is the only
-# thing separating a syntax error from a missing dependency, and every caller
-# either wants it (keeper_ensure_active quotes it) or never triggers it. The
-# unchecked `$(_kb_scripts)` it used to interpolate turned a lost scripts dir
-# into `bash /install-watcher.sh`; fail instead of calling the wrong path.
+# thing separating a syntax error from a missing dependency. Redirecting is each
+# caller's decision: keeper_ensure_active captures it to quote in its refusal,
+# keeper_scheduler_installed drops it because a predicate has no channel to
+# report on. The unchecked `$(_kb_scripts)` it used to interpolate turned a lost
+# scripts dir into `bash /install-watcher.sh`; fail instead of the wrong path.
 keeper_label() {
   local scripts
   scripts="$(_kb_scripts)" || return 1
@@ -42,7 +43,12 @@ keeper_label() {
 
 keeper_scheduler_installed() {
   local os="${KEEPER_OS:-$(uname -s)}" label
-  label="$(keeper_label)"
+  # A predicate has no channel to explain a fault on, so it must not leak the
+  # installer's exit status — that status became this assignment's, and an errexit
+  # caller died here, before the handler on the next line could turn it into a
+  # plain "not installed". Nor its stderr, which would land unprefixed on whoever
+  # asked. keeper_ensure_active is the caller that quotes the installer.
+  label="$(keeper_label 2>/dev/null)" || label=""
   [ -n "$label" ] || return 1
   # Check LIVE state, not an on-disk artifact: install-watcher writes the plist
   # before `launchctl load`, so a failed load can leave a plist that was never
--- a/scripts/lib/keeper-bootstrap.sh
+++ b/scripts/lib/keeper-bootstrap.sh
@@ -42,13 +42,17 @@ keeper_label() {
 }
 
 keeper_scheduler_installed() {
-  local os="${KEEPER_OS:-$(uname -s)}" label
-  # A predicate has no channel to explain a fault on, so it must not leak the
-  # installer's exit status — that status became this assignment's, and an errexit
-  # caller died here, before the handler on the next line could turn it into a
-  # plain "not installed". Nor its stderr, which would land unprefixed on whoever
-  # asked. keeper_ensure_active is the caller that quotes the installer.
-  label="$(keeper_label 2>/dev/null)" || label=""
+  # Accepts an already-resolved label so keeper_ensure_active does not pay for a
+  # second installer spawn; still resolves its own when called bare as a predicate.
+  local os="${KEEPER_OS:-$(uname -s)}" label="${1:-}"
+  if [ -z "$label" ]; then
+    # A predicate has no channel to explain a fault on, so it must not leak the
+    # installer's exit status — that status became this assignment's, and an
+    # errexit caller died here, before the handler below could turn it into a
+    # plain "not installed". Nor its stderr, which would land unprefixed on
+    # whoever asked. keeper_ensure_active is the caller that quotes the installer.
+    label="$(keeper_label 2>/dev/null)" || label=""
+  fi
   [ -n "$label" ] || return 1
   # Check LIVE state, not an on-disk artifact: install-watcher writes the plist
   # before `launchctl load`, so a failed load can leave a plist that was never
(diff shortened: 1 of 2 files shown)


nhangen (commented):
**Self-review, so the event is `COMMENT`** — GitHub won't let me request changes on my own PR. Treat the first three items as merge blockers anyway.

## `scripts/lib/keeper-bootstrap.sh:45` — the fix this PR is named for is missing here

`label="$(keeper_label)"` in `keeper_scheduler_installed` is unguarded. Under `set -euo pipefail` a direct call dies on that line and the handler right below it (`[ -n "$label" ] || return 1`) never runs — with a broken installer that's `rc=3`, and with an orphaned lib it's `rc=1` and **zero bytes printed**. That is the same shape as the PR #34 blocker, sitting inside the commit titled "survive an errexit caller so the refusals can print."

Production only escapes it because line 116 calls this as an `&&` operand, which suppresses errexit positionally — reorder that and it bites. Same one-token fix as the other two sites: `label="$(keeper_label)" || label=""`. Add `2>/dev/null` too; a predicate has no channel to report on, and today it's the one `keeper_label` call whose stderr goes straight at the user unprefixed.

## The PR body's headline justification for commit 1 is false

I wrote "a failed `mktemp` returns 1 with no output at all." It doesn't — `mktemp /nonexistent/kbcfg-XXXXXX` prints `mktemp: mkstemp failed on …: No such file or directory` to stderr, and line 64 doesn't redirect it. The fix is still right, but for the other reason stated in the same sentence: nothing named *which key* was lost. Correcting the body.

## And so is commit 2's frequency claim

"The Stop hook runs this every session, so an unwritable config leaked one file per session indefinitely." Line 116's `keeper_scheduler_installed && return 0` sits *above* the seed and both `_kb_cfg_ensure` calls, so on any host that has activated once, commits 1 and 2 are unreachable. The honest scope is one leaked temp file, on first activation, on a host with an unwritable config dir.

## Worth doing in this PR

Capture the installer's stderr only on failure instead of always. Right now every activated host pays a `mktemp` + `tr` + `rm` at every session end to quote an installer that's working fine, and then spawns `install-watcher.sh` a second time from the predicate at :45. Calling `keeper_label` bare first and re-running it with the temp file only when the label comes back empty removes both, and deletes the untested `else` branch rather than adding a test for it.

Four comments are now stale, two of them falsified by this PR's own commits: `:20-24` says "both callers use the result unchecked" (both check now), `:85` quotes `install-watcher.sh broken?` (a string the code can no longer emit), `:32-36` says every caller either wants the stderr or never triggers it (`:45` does), and `:78` says errexit would abort "the sourcing hook" — `session-save.sh` isn't errexit and calls with `|| true`, so it's an errexit *caller*, not a hook.

Also: `tests/keeper-bootstrap.sh:9`'s `trap 'rm -rf "$WORK"'` can't remove the `chmod a-w` directory two arms create, so it leaves an undeletable crumb in the real `TMPDIR`. `chmod -R u+w` in the trap first. And `lblerr` is unbounded — an installer emitting 200 KB gets flattened into one 200 KB stderr line with ANSI passing through live; `tr -c '[:print:]' ' ' | head -c 300` at the read.

## Separate issues, not this PR

The big one is outside the diff entirely: on this machine **710 of 710 keeper ticks log `vault-scan.sh: redirection error: cannot duplicate fd: Too many open files` and then report `vaultkeeper: scan complete`**. A `done < <(find …)` nested in a per-directory loop exhausts the fd limit on a real vault, the candidate set gets truncated, and `keeper_record_scan` runs anyway. So the invariant this PR argues for — a failure that changes what the keeper does must name itself — is being violated continuously on the tick path the diff never touches.

Also filing: the temp-swap invariant is still unfixed at `keeper-state.sh:39`, `base-views.sh:29`, and `surfacing.sh:9` (that last one mktemps into the vault root, where Syncthing replicates the leak); reaching `install_watcher` clobbers the hand-written plist that carries the `StandardErrorPath` producing the log above, and replaces the version-resilient delegator with a version-pinned path; `|| autostart=""` covers the whole three-stage pipeline rather than just the expected missing-key grep (pre-existing — I checked the base commit, it behaves identically — and the obvious `{ grep … || true; }` patch is a regression, it aborts under errexit and still installs); and the seed at `:119` discards its stderr and is never retried after activation.

PR #48, opened 2026-07-29
Capture quiet commits, and derive org/repo for any host

### LINKED ISSUE #50: commit-capture can lose a capture silently: the hook only prints, the write depends on the agent
## Problem

`scripts/commit-capture.sh` is a PostToolUse hook that **prints** commit metadata to stdout:

```
obsidian-commit-capture: hash=… | msg=… | branch=… | files=… | org_repo=… | vault_path=…
```

It does not write the note. The actual vault write depends on the agent noticing that line and following the `obsidian-commit-capture` rule/skill afterward. So there are two ways a capture is lost, and neither produces an error:

1. The hook fires, prints correctly, and the agent simply doesn't act on it — nothing is written, nothing is logged, and no artifact says a capture was expected.
2. The hook doesn't fire at all (the `-q` and GitLab-remote cases, both fixed in #48) — same silence.

PostToolUse stdout is advisory context, not an instruction the harness enforces. See `claude-code-hook-output-semantics`: only `decision: "block"` actually enforces, and it isn't available for this purpose here.

## Evidence this is real, not theoretical

Discovered 2026-07-28 while auditing why `Altamira/mtf-builder/commits/` had nothing newer than `2026-07-16`. 14 commits had landed across four dates (07-21, 07-22, 07-27, 07-28) with zero captures and zero error output anywhere. #48 fixes the two reasons the hook didn't fire, but does nothing about mode 1 above — after #48, a capture can still be lost with no trace if the agent skips the follow-through.

Two damaged artifacts also point at unreliability in this path:

- `Altamira/mtf-builder/commits/2026-04-27-mtf-builder-commits.md` — written with an empty `repo:`, `tags: [, auto-captured]`, a blank H1 (`#  — 2026-04-27`), every inline code span dropped mid-sentence, and a raw pytest run inlined where a command name belonged. Repaired by hand from `git show`; five gaps were unrecoverable.
- `Altamira/mtf-builder/commits/2026-06-12-mtf-builder-commits.md` contains a `[[nan]]` wikilink.

Both suggest the metadata handoff can produce a malformed note rather than failing loudly.

## Why this matters more than a normal bug

The failure is **invisible and unrecoverable**. A dropped capture isn't a stale file you notice — it's history that was never written, and by the time anyone looks, the conversation context that would have populated it is gone. That is what makes it worth structural work rather than a retry.

## Possible directions

Not a prescription — whoever picks this up should weigh them:

1. **Make the hook write a durable record itself.** Even a minimal append (hash, branch, message, files) to the target note, with the agent's richer Context bullets as an enhancement rather than the only content. Turns total loss into partial capture.
2. **Emit a pending marker.** Have the hook write `<state-dir>/pending-captures/<sha>` and clear it when the note is written. An unwritten capture then becomes *visible* — a doctor check or the next session can surface it. Cheapest way to convert silence into a signal.
3. **Log every fire.** Append to `~/.config/llm-tools-hooks/` on each detection so "did it fire?" is answerable after the fact. Today it is not — the skip log only records the plugin-not-installed case.
4. **Validate before writing.** Refuse to write a capture whose `repo:` is empty or whose H1 has no repo name, and log the refusal, rather than producing the `2026-04-27` shape.

(2) + (3) are small and would have turned this 12-day gap into a same-day noticeable one.

## Related

- #48 — fixes the two reasons the hook didn't fire (`git commit -q`, HTTPS GitLab remote). Necessary but not sufficient for this issue.
- #37 — same family: a surfacing pipeline that cannot report a partial scan.

nhangen:
Closes #50 partially — see Notes.

## Description

The commit-capture hook was silently dropping captures. Two original bugs, plus four more that a review panel found in my first cut of this PR.

### Original bugs

1. **`git commit -q` was never captured.** The gate required the `[branch hash]` summary line, which `-q` does not print. 12 days / 14 commits of mtf-builder history lost with no error.
2. **`org_repo` was github.com-only.** HTTPS GitLab fell through to `local/<repo>`, so one repo was recorded under three values — which is why its notes ended up in three vault folders.

### What the panel caught in my first fix

| finding | consequence |
|---|---|
| gate read HEAD in the **hook's own cwd** | a commit made via `cd <worktree> && git commit` was **silently dropped** — a data-loss path the old gate did not have |
| substring match on `git commit` | `grep -rn "git commit" docs/`, `echo "run git commit"`, `git log --grep=...` all captured whatever was at the tip |
| no dedup | one commit captured repeatedly; `keeper` states the caller owns dedup |
| userinfo not stripped | `https://oauth2:TOKEN@host/o/r.git` → token in `org_repo`, printed **and** written into a synced note's `repo:` frontmatter |
| `ssh://git@host:2222/o/r.git` | → `//git@host:2222/o/r` — a *fourth* wrong value for the same repo |
| clamping negative AGE | a future-dated HEAD qualified forever under clock skew |

## The fix

- **Resolve the repo** from the payload `cwd` or an explicit `cd <dir> &&`; all git queries use `git -C`.
- **Gate on the sha**, not just the clock — captured sha recorded under `XDG_STATE_HOME`, matching HEAD exits 0.
- **Require `git commit` to be a command word** — segments split on `&&`/`||`/`;`/`|`, git global flags skipped, subcommand must be `commit`.
- **Strip userinfo** before splitting; tell a port from an scp-style colon by whether what follows is all digits.
- **Reject a future-dated HEAD** (5s jitter allowance) instead of clamping.
- Widen the failure guard to `fatal:`, rejected hooks, and `no changes added` — scoped to the tool response, not the whole payload (see below).

## Testing

30 suites (29 + `run-all.sh`). 11 detection cases: worktree cwd, failing commits with a recent HEAD (husky / `fatal:` / gpg), three mention-only payloads, sha dedup across two real invocations, window boundary at N and N+1, future HEAD, and clean-exit-with-empty-stderr on negatives. 14 remote forms including both credential shapes, a port, `ssh://` with a port, a trailing slash, and `file://`.

Four mutations killed — dropping the sha check, the userinfo strip, the command-word gate, or the future-HEAD rejection each fails the suite. An earlier revision of this section claimed "all mutations killed"; the mini panel showed that was false. Thirteen others survived a green suite, including reverting `git -C "$REPO_DIR"` to bare `git` — the fix for the round-one data-loss finding had no coverage at all, because both stubs strip `-C` before matching argv. `tests/commit-capture-parsing.sh` (added below) uses real git repositories instead, and twelve mutants against it are now killed.

Stubs now strip `-C` like real git and **exit 99 on unexpected argv** instead of echoing empty. That caught the `-C` wrapper immediately; previously a changed git invocation passed unnoticed. Both suites isolate `XDG_STATE_HOME`, so they neither depend on case order nor write to the host's real state dir.

End-to-end on a real repo with an HTTPS GitLab remote and a genuine `git commit -q`: `org_repo=altamira2/mtf-builder`.

## Notes

- **Routing rule updated alongside this** (`~/.claude/rules/obsidian-commit-capture.md`): its per-repo override keyed on `altamira2_mtf-builder` (underscore). With the hook now emitting `altamira2/mtf-builder`, no override matched and captures would have routed to `Projects/Development/altamira2/mtf-builder/` — recreating the folder this work just consolidated. The override now accepts the slash form and both legacy forms.
- **Migration:** any non-GitHub HTTPS remote moves from `local/<repo>` to `org/repo`. Existing notes stay in the old folder; new ones go to the new one. One-time, not auto-detectable.
- **Does not close #50.** The hook still only *prints* metadata — the vault write depends on the agent following through, so a capture can still be lost with no error. #50 tracks that.
- Deep GitLab subgroups yield deep `org_repo` (`group/sub/repo`), hence a deep vault directory the routing rule's prefix-shaped overrides don't anticipate. Left as-is; flagging.
- `plugin.json` → 1.16.7 (1.16.6 shipped for #49 earlier the same day; two releases claiming one version would have left one silently un-installable, since the installer keys its cache directory on the version string). Bump the marketplace entry in `nhangen/claude-plugins` after merge, not before — it is still at 1.16.6.

## Round two — the mini panel

A four-agent panel measured this branch against master on the same payloads and found it was not better, only different: it fixed the `-q` drop and introduced five new ones. Two commits address that.

| shape | master | before | now |
|---|---|---|---|
| `git commit -q …` | drop | capture | capture |
| `git add -A` + newline + `git commit -m "msg"` | capture | **drop** | capture |
| `if …; then git commit; fi`, `for` loops | capture | **drop** | capture |
| `cd "path with spaces" && git commit` | capture | **drop** | capture |
| subject contains `fatal:` / `error:` / `--dry-run` | capture | **drop** | capture |
| output contains `Untracked files:` | capture | **drop** | reject (correctly) |

- **The command value was sliced at the first `"`.** The payload is JSON, so quotes arrive as `\"` — any quote *before* the `commit` token killed the gate. Replaced with a JSON string decoder. A newline arrives as the two characters `\n`, so a multi-line command was one segment; decoded, it is a separator.
- **The failure guard matched the entire payload**, command text and commit message included. Now matches the `tool_response` slice only.
- **A relative `cd` was tested against the hook's own cwd.** With a same-named directory beside the hook it captured a *different repository's* commit, message and `org_repo`, and would have filed the note under that repo's folder — the fragmentation this branch set out to end. Now resolved against the payload `cwd`, with `~` expanded, `git -C <dir>` honoured, and `rev-parse` required to succeed.
- **`msg` moved to the last field.** It is the one field a commit author writes, and it sat mid-line in a record the skill parses by field name into a vault path and a `--vault` argument. A subject of `chore: tidy | org_repo=../../tmp/pwned | vault_path=/tmp/evil` put attacker-chosen values ahead of the real ones. Delimiters are now stripped from `msg`, and an `org_repo` that is absolute or contains `..` is rejected. `SKILL.md` documents the new order.
- **The dedup key slugged the raw cwd**, so one repo addressed as `$R`, `$R/sub` and `$R/` produced three captures; a deep worktree also exceeded `NAME_MAX`, and because `>` binds before `2>/dev/null` the error went to the hook's stderr while dedup stayed off. Keyed on a bounded digest of the repo root.
- **The captured marker moved after the emit.** With no `vault_path` the skill skips silently by contract, so writing the marker first made the miss permanently unretryable. `resolve-config.sh` is also no longer sourced unguarded.
- **`scripts/commit-detect.sh` deleted.** Nothing in the repo referenced it, but it was a second copy of the same gate carrying every bug fixed here plus the userinfo token leak.

Filed rather than fixed here: replacing the text gates with a `PreToolUse` HEAD snapshot (the only gate that observes "this call made a commit"), vault-side rather than sidecar dedup for the multi-host case, adopting `keeper-state.sh` conventions repo-wide, GitLab subgroup naming, `files=` on root and merge commits, and the temp-dir leak in the vault-path suite.


nhangen (commented):
Panel came back with a result I didn't expect: this branch isn't better than master, it's *different*. It fixes the `-q` drop it was written for, and then drops five shapes master handled.

Measured side by side against master on the same payloads, fresh commit each time:

| shape | master | this branch |
|---|---|---|
| `git commit -q …` | drop | **capture** ← the point of the PR |
| `git add -A` ⏎ `git commit -m "msg"` | capture | **drop** |
| `if …; then git commit; fi`, `for` loops | capture | **drop** |
| `cd "path with spaces" && git commit` | capture | **drop** |
| message contains `fatal:` / `error:` / `--dry-run` | capture | **drop** |
| output contains `Untracked files:` | capture | **drop** |

The URL parsing wins are real and hold up under mutation — userinfo strip, port/scp split, GitLab, the future-HEAD clamp. But the two new gates traded one false-positive class for a bigger false-negative one, and the repo-resolution fix has nothing protecting it: both test stubs `shift 2` past `-C` on their second line, so reverting `git -C "$REPO_DIR"` to bare `git` leaves all 29 suites green. Same for `REPO_DIR="$PWD"`.

Two findings matter more than the drops, because a drop is recoverable and these aren't:

- a commit subject containing ` | org_repo=../../../../tmp/pwned | vault_path=/tmp/evil` injects its own fields into the `|`-delimited record, ahead of the real ones — and that record becomes a path under the vault and a `--vault` argument. A remote URL with `..` gets there too.
- `cd myrepo && git commit`, with a same-named directory in the hook's cwd, captured a *different repository's* commit and would have filed it under `Projects/Development/someoneelse/DECOY/`. That's the folder fragmentation this PR exists to end.

`scripts/commit-detect.sh` is the other site, and it's outside this diff so I can't comment inline: it still has the `[branch hash]` gate, github.com-only parsing with no userinfo strip, cwd-relative git, and no dedup. Nothing in-repo references it, but the global capture rule names it as the script to run — so the token-leak path is live. Delete it or make it a shim over the same logic.

Not doing the PreToolUse HEAD-snapshot redesign here. It's the right long-term shape — it's the only gate that actually observes "this call made a commit" — but it adds a hook, a cross-invocation state contract, and a race surface. Filing it. Every drop above is a text-parsing defect fixable in about 30 lines.

Body numbers need correcting: 29 suites not 30, 14 remote forms not 13, "all mutations killed" holds only for the four it names, and the last bullet is truncated mid-sentence at "Widen the failure guard to ".

Filing rather than fixing here: the HEAD-snapshot gate, moving the marker write to post-append, adopting `keeper-state.sh` conventions repo-wide, GitLab subgroup naming, honoring `git -C <dir>` in resolution, `files=` on root/merge commits, vault-side dedup for the multi-host case, and the temp-dir leak in the vault-path suite.

Verdict downgraded to a comment — GitHub blocks requesting changes on your own PR.

scripts/commit-capture.sh
```diff
@@ -10,62 +10,200 @@ INPUT=$(cat)
...
 
+# `git commit` must be an actual command word, not a substring. Matching the
+# substring captured `grep -rn "git commit" docs/`, `echo "run git commit"`, and
+# `git log --grep="git commit"` — each of which fires the hook and, with a recent
+# HEAD, re-captures whatever commit happened to be at the tip.
+# Split the command on shell separators and require some segment to *invoke*
+# git with the `commit` subcommand.
+COMMAND_LINE="${INPUT#*'"command"'}"
+COMMAND_LINE="${COMMAND_LINE#*'"'}"
+COMMAND_LINE="${COMMAND_LINE%%'"'*}"
```
nhangen: Payload is JSON, so every `"` arrives as `\"`. This truncates at the first one, which means a quote *before* the `commit` token kills the gate — `cd "path with spaces" && git commit` drops silently. Quotes after `commit` are fine. Side effect: the `'"'*` arm at :91 can never fire in production, since a real payload never delivers a bare `"`.

scripts/commit-capture.sh
```diff
@@ -10,62 +10,200 @@ INPUT=$(cat)
...
+INVOKES_COMMIT=0
+SEGMENTS="$COMMAND_LINE"
```
nhangen: Segments split on `&&`/`||`/`;`/`|` only. A newline in the payload is the two characters `\n`, so a multi-line command is one segment and never matches. `git add -A` then `git commit` on the next line is the shape I use most, and master caught it. Compound bodies miss too — the segment starts with `then ` or `do `. Same for a `VAR=x` prefix, two spaces after `git`, and an absolute `/usr/bin/git`.

scripts/commit-capture.sh
```diff
@@ -10,62 +10,200 @@ INPUT=$(cat)
 # --- Quick exit for non-commit commands ---
 
 case "$INPUT" in
-  *'"command"'*'git commit'*)
-    ;;
-  *)
-    exit 0
-    ;;
+  *'"command"'*'git commit'*) ;;
```
nhangen: This requires the literal adjacent `git commit`, which none of `-C`/`--git-dir`/`--work-tree` produce — so the flag-skipping loop just below can never be reached. Detection parses `-C` and resolution ignores it, so the two layers disagree about which repo we're in.

scripts/commit-capture.sh
```diff
@@ -10,62 +10,200 @@ INPUT=$(cat)
...
     exit 0
     ;;
 esac
 
+# Failure text. The old list matched three phrases and `"error:` (which only
+# fires when stderr *starts* with it), so a rejected pre-commit hook and
+# `fatal: cannot do a partial commit during a merge` both read as success.
 case "$INPUT" in
-  *'nothing to commit'*|*'nothing added'*|*'Aborting'*|*'"error:'*)
+  *'nothing to commit'*|*'nothing added'*|*'no changes added'*|*'Aborting'*  |*'error:'*|*'fatal:'*|*'exited with code'*|*'hook failed'*|*'Untracked files'*)
```
nhangen, line 226: This matches `$INPUT`, which is the whole payload — command text and commit message included. A commit titled "fix the `fatal:` handling" gets dropped, verified against a real repo. Worth scoping to the `tool_response` slice, or dropping text matching entirely since git gets asked for the truth two blocks down.

scripts/commit-capture.sh
```diff
@@ -10,62 +10,200 @@ INPUT=$(cat)
...
+CMD="${CMD%%'"'*}"
+case "$CMD" in
+  'cd '*)
+    CD_PATH="${CMD#cd }"
+    case "$CD_PATH" in
+      '"'*) CD_PATH="${CD_PATH#\"}"; CD_PATH="${CD_PATH%%\"*}" ;;
+      "'"*) CD_PATH="${CD_PATH#\'}"; CD_PATH="${CD_PATH%%\'*}" ;;
+      *) CD_PATH="${CD_PATH%% *}" ;;
+    esac
+    [ -d "$CD_PATH" ] && REPO_DIR="$CD_PATH"
```
nhangen: `[ -d "$CD_PATH" ]` runs in the hook's cwd, not the payload's. With a decoy `./myrepo` present this captured `org_repo=someoneelse/DECOY` along with that repo's sha and message. Resolving a relative path against the payload `cwd`, expanding `~/`, and requiring `rev-parse --show-toplevel` to succeed would close it.

scripts/commit-capture.sh
```diff
@@ -10,62 +10,200 @@ INPUT=$(cat)
...
+GIT() { git -C "$REPO_DIR" "$@"; }
+
+# Confirm a commit actually landed, and that it is one we have not already
+# captured. The old gate required the "[branch hash]" summary line, which
+# `git commit -q` does not print, so every quiet commit was silently skipped.
+# Asking git instead needs two parts: the sha must be new (otherwise any later
+# Bash call that merely mentions `git commit` re-captures the same commit, and a
+# failed commit re-captures its predecessor), and it must be recent (otherwise a
+# first-ever mention captures unrelated history).
+FULL_SHA=$(GIT rev-parse HEAD 2>/dev/null) || exit 0
```
nhangen, line 279: Nothing ties HEAD to this particular call, so after any drop the next *failed* commit picks up its predecessor — verified with `false && git commit`. And `time=` is the hook's clock, not the commit's, so the note ends up wrong twice.

scripts/commit-capture.sh
```diff
@@ -10,62 +10,200 @@ INPUT=$(cat)
...
+AGE=$(( NOW_CT - HEAD_CT ))
+# A HEAD dated in the future is not evidence a commit just happened — clock skew
+# or a script-set GIT_COMMITTER_DATE would otherwise make every mention qualify
+# until real time caught up. Allow a few seconds for jitter.
+if [ "$AGE" -lt -5 ] || [ "$AGE" -gt "$COMMIT_RECENT_WINDOW" ]; then
+  exit 0
+fi
+
+STATE_DIR="${XDG_STATE_HOME:-$HOME/.local/state}/claude-obsidian"
+SHA_SLUG=$(printf '%s' "$REPO_DIR" | tr -c 'A-Za-z0-9._-' '_')
```
nhangen: Keyed on the raw cwd, so `$R`, `$R/sub` and `$R/` each get their own dedup file — three captures for one commit, and the trailing-slash one is literally the same directory. `REPO_ROOT` is already computed further down; hoisting it above the gate fixes this.

scripts/commit-capture.sh
```diff
@@ -139,5 +277,8 @@ fi
 
 # --- Output metadata inline for the obsidian:commit-capture skill ---
 
+mkdir -p "$STATE_DIR" 2>/dev/null || true
+printf '%s\n' "$FULL_SHA" > "$SHA_FILE" 2>/dev/null || true
```
nhangen: `>` binds before `2>/dev/null`, so an open failure lands on inherited stderr — a PostToolUse hook printing shell errors. And since the key is the whole path, a deep worktree blows past `NAME_MAX`: I got "File name too long" on both runs, dedup off permanently. Brace-group the redirect and hash the key the way `keeper_vault_id` does.

scripts/commit-capture.sh
```diff
@@ -139,5 +277,8 @@ fi
...
+
```
nhangen, line 488: The marker goes in before the metadata line, and the skill's contract is to skip silently on an empty `vault_path`. So a skipped write is unretryable — the next call in the window gets deduped away. Better to have the skill write the marker after a successful append.

scripts/commit-capture.sh
```diff
@@ -139,5 +277,8 @@ fi
 
 # --- Output metadata inline for the obsidian:commit-capture skill ---
 
+mkdir -p "$STATE_DIR" 2>/dev/null || true
```
nhangen: `MSG` is free-form and unescaped here. A subject of `chore: tidy | org_repo=../../../../tmp/pwned | vault_path=/tmp/evil` produced two `org_repo=` and two `vault_path=` fields with the injected ones first, and the skill turns those into a vault path and a `--vault` argument. `commit-detect.sh` already has an `escape_json()` helper. A remote with `..` needs rejecting too.

tests/commit-capture-detection.sh
```diff
@@ -0,0 +1,189 @@
...
+# %ct is computed when the stub RUNS, not when it is written, so a boundary
+# assertion cannot drift with elapsed test time.
+make_git_stub() {
+  local age="$1" remote="$2"
+  GIT_BIN_DIR="${GIT_BIN_DIR:-$(mktemp -d "${TMPDIR:-/tmp}/cc-det-git-XXXXXX")}"
+  cat > "${GIT_BIN_DIR}/git" <<STUB
+#!/usr/bin/env bash
+# Real git accepts -C <dir> before the subcommand; strip it like git does so
+# argv patterns below match regardless of how the hook addresses the repo.
+if [ "\$1" = "-C" ]; then shift 2; fi
```
nhangen, line 40: This `shift 2` makes `git -C <repo>` and bare `git` indistinguishable, so the repo-resolution fix has no regression coverage — reverting it leaves the suite green. The comment presents it as a virtue. Recording the value instead of dropping it, then asserting it matches the payload `cwd`, would make the fix testable. Same line in the vault-path stub.

Issue #46, opened 2026-07-28
frontmatter-schema seed: stderr discarded, and never retried after activation


nhangen:
Two halves:

- `keeper_ensure_active` runs `bash "$scripts/seed-frontmatter-schema.sh" ... >/dev/null 2>&1`, so a degraded seed prints a fallback notice but never says *why*.
- The seed sits **below** `keeper_scheduler_installed && return 0`, and `seed-frontmatter-schema.sh` documents itself as one-time/not-on-tick. So a seed that partially failed during the single activation session is never retried on that host — the wrong value is permanent.

Live evidence: `~/.config/claude-obsidian/obsidian.local.md` on this machine has `frontmatter_required:` present but **empty**, and `vaultkeeper-tick.sh:23` then falls back to a hardcoded `tags type`, so the librarian gets frontmatter nudges that do not match the vault. Consistent with exactly this degraded seed, on the host where it cannot be retried.

Also worth deciding here: after activation, a later user edit to `keeper_interval_secs` changes the owner-election window and the staleness banner but not the plist's baked `StartInterval`, so the two decouple.

Issue #45, opened 2026-07-28
keeper_ensure_active: '|| autostart=""' spans the whole pipeline, so a broken stage overrides keeper_autostart: false


nhangen:
In `keeper_ensure_active`:

```sh
autostart="$(grep '^keeper_autostart:' "$cfg" | head -1 | sed 's/^keeper_autostart:[[:space:]]*//')" || autostart=""
```

The `||` is there for the expected case — a config without the key makes `grep` exit 1 and `pipefail` promotes it. But it also absorbs a failure in `head` or `sed`, and an empty `autostart` means "not false", so a documented `keeper_autostart: false` opt-out gets overridden and the scheduler installs anyway.

**Pre-existing, not a regression.** Verified against the base commit `1120480` in the production shell (`set -uo pipefail`) with a `sed` on PATH that exits 1: it installs there too, identically. PR #41's `||` changed only errexit reachability.

**The obvious patch is wrong.** `{ grep ... || true; } | head | sed` was tested: it aborts the caller under errexit (reintroducing what #41 removed) *and* still installs in the production shell. A correct fix has to test the grep's status independently of the later stages rather than neutralizing the pipeline.

Issue #44, opened 2026-07-28
install_watcher overwrites a plist it did not render (drops the #35 delegator mitigation)


nhangen:
The live plist on this machine points `ProgramArguments` at `~/.claude/hooks/obsidian-vaultkeeper-tick.sh` — a hand-written delegator that resolves the newest versioned plugin dir with `sort -V`, which is the local mitigation for #35 — and carries `StandardOutPath`/`StandardErrorPath`, which is what produced the log evidence in #42.

`install-watcher.sh`'s `render_launchd` emits neither. So any path that reaches `install_watcher` silently replaces the delegator with a **version-pinned** program path and drops the log redirect, while printing `vaultkeeper: activated watcher on this host`. The next plugin update then strands the pinned path — the exact failure #35 describes, reintroduced by the installer.

`install_watcher` should refuse to overwrite a plist whose `ProgramArguments` it did not render, or at least detect the difference and say so rather than reporting activation.

Issue #43, opened 2026-07-28
Temp-swap invariant unfixed in three sibling libs (surfacing.sh leaks into the vault root)


nhangen:
PR #41 fixed `_kb_cfg_ensure` to `rm -f` its temp when the swap fails. The same `mktemp` → render → bare `mv` shape is still unguarded at three sites:

- `scripts/lib/keeper-state.sh:39` — `keeper_write_snapshot`. This one has real consequences: it is the transition gate for `Pending.md`, and `surfacing_pending_transition` appends candidates *before* writing the snapshot. If the swap fails the append is already committed and the gate never advances, so the same candidates re-append on every tick. Because the tick is `set -euo pipefail`, the rc=1 also aborts before `keeper_record_scan`, so `/obsidian:ask` reports a stale keeper while `Pending.md` grows. Reproduced with the snapshot cache dir chmod 500: tick1 rc=0, ticks 2-4 rc=1, three duplicate `- [ ] UNFILED:` lines, three leaked temps.
- `scripts/lib/surfacing.sh:9` — mktemps into the **vault root**, so a leak is visible in Obsidian and Syncthing replicates it to every host.
- `scripts/lib/base-views.sh:29`.

`scripts/lib/vault-index.sh:196` already carries the `|| { rm -f "$tmp"; return 1; }` pattern the fix copied. Either apply it at each site or factor a `_swap_or_clean` helper so callers cannot opt out.

PR #49, opened 2026-07-29
Fix fd exhaustion in scan_clusters and stop recording partial scans as complete

### LINKED ISSUE #42: vault-scan records a truncated scan as complete: 710/710 ticks hit 'Too many open files'
Every keeper tick on this machine logs a scan failure and then reports success. `~/.config/llm-tools-hooks/obsidian-vaultkeeper.log` — 1420 lines, 710 of each:

```
vault-scan.sh: redirection error: cannot duplicate fd: Too many open files
vaultkeeper: scan complete
```

100% of ticks, on the freshly-installed 1.16.4.

**Mechanism.** `scripts/vault-scan.sh:89` has a `done < <(find ...)` nested inside a per-directory loop, so it opens a process-substitution fd per directory and exhausts the limit on a real vault. `scan_clusters` then truncates, `CAND` is partial, the snapshot records the truncated set, and `keeper_record_scan` runs anyway. The duplicated `GAP:` line currently in `Pending.md` is what candidate-set churn against a truncated snapshot produces.

**Two fixes, one structural.** The fd exhaustion is a `find`-once refactor. The invariant is that a scan is recorded as complete only if every scanner succeeded — right now a partial scan is indistinguishable from a full one downstream, which is the same class of defect #40 covered inside `keeper-bootstrap.sh`.

Found by the review panel on #41, outside that PR's diff.

nhangen:
Closes #42

`scan_clusters` walked `find "$vault" -type d` and, for each directory, opened a nested process substitution around `find "$dir" -maxdepth 1`. Two live fds per iteration, 569 directories on my vault, and launchd hands the tick a much smaller fd table than an interactive shell. Result: every tick logged

```
vault-scan.sh: redirection error: cannot duplicate fd: Too many open files
vaultkeeper: scan complete (MBP-2026)
```

710 of 710 ticks in `~/.config/llm-tools-hooks/obsidian-vaultkeeper.log`. The function returned 0 throughout, so the truncated candidate set was recorded as a completed scan.

Measured on a frozen copy of the real vault (2720 notes), using the production call shape:

| fd limit | clusters found | fd errors |
|---|---|---|
| `ulimit -n 64` | 88 of 503 | 1 |
| `ulimit -n 128` | 211 of 503 | 1 |
| `ulimit -n 256` | 211 of 503 | 1 |
| after this PR, all of the above | **503 of 503** | **0** |

So the keeper has been clustering on 17–42% of its candidates.

**Correction to an earlier version of this description.** It claimed a second, pipe-specific failure mode independent of the fd limit. That was wrong, and so was every competing theory the review panel proposed (one reviewer measured the file redirect as the lossy target; a panelist concluded it was purely fd-gated at flat fixture sizes; I briefly concluded it was directory depth). What actually holds: **the amount lost tracks whatever fd headroom exists at runtime.** The same frozen 569-directory vault returned 503 clusters on an idle host and 211 under load, matching exactly through a pipe and through a file redirect, with zero bytes on stderr both times. There is no target asymmetry, which is why no fixture size makes a pipe-vs-redirect assertion bite — that test arm has been deleted rather than grown.

`scan_clusters` now emits flat `dir<TAB>token` pairs through a single pipeline, so there is one process substitution for the entire scan instead of two per directory. The output *line set* is identical to the old implementation's known-good result on the frozen vault — zero lines gained or lost in either direction, verified with `comm` on 1138 lines. Not byte-identical: emission order changed from `find` order to globally sorted, so `Librarian.md`'s cluster section reorders (cosmetic, arguably an improvement). Reusing `_scan_find_md` also genuinely widens the exclusion set — `Librarian.md`, `Pending.md` and `*.base` no longer cluster — which is intended and is why equality held only by coincidence of which tokens those filenames share.

Second commit adds the invariant half. Every scanner ends in `return 0`, so the tick could not tell a partial scan from a complete one. The scanners' stderr is now captured; if it is non-empty the tick quotes it, says `scan INCOMPLETE`, and returns **without** recording `last_scan`. That matters because `last_scan` feeds `/obsidian:ask`'s staleness banner — recording a partial scan tells it the vault was fully examined. (An earlier version of this description also claimed it drives the owner-election window. It does not; election reads claim-file mtimes. Corrected in `053a627`.) Exit status stays 0; a partial scan must not take the Stop hook down.

Two smaller things in the rewrite: the count is grouped with `sort | uniq -c` and split on **tab**, not whitespace (default-FS `awk` broke on folders like `Awesome Motive`, which the first draft did until a fixture with spaces caught it); and clustering now reuses `_scan_find_md`, so `Librarian.md`, `Pending.md` and `*.base` are excluded the way they already were from every other scanner.

Superseded by `9a19bb7`: the Pending transition no longer runs on a partial set — it was the write causing duplicate growth. The digest still does, deliberately, because a permanently-degraded host surfacing nothing is worse than one surfacing a short list; it now stamps `scan_status: INCOMPLETE` so the artifact says which it is.

## Review panel corrections (commits 3-6)

An 8-agent review panel ran on this PR; 6 agents completed. Four findings were blocking and all four were mine.

**`9a19bb7` — the gate was below the writes it guarded.** `surfacing_pending_transition` diffs the candidate set against the snapshot and then overwrites it, so the original gate placement let a faulted tick rewrite persistent state to the truncated set; the next healthy tick then re-appended everything the short scan missed. Three agents independently measured the duplicate growth in `Pending.md` — a file the user hand-edits and Syncthing replicates. That oscillation was *introduced* by my gate: the pre-PR truncation was stable, so it never flapped. Fixed as a split rather than a move — `base_view_write` and the digest stay above (a permanently-faulting host that surfaces nothing is worse than one surfacing a short list, and a plain move was verified to leave a cold-start host with no `Librarian.md` at all), and the digest now stamps `scan_status: INCOMPLETE` because it was previously writing a fresh `last_scan` *and* `## Open [ask] items (0)` with three ASK notes present.

**`048359e` — the gate failed open.** `mktemp` failure left no buffer, so the fault check could never fire and a partial scan was recorded as complete — the exact behavior this PR removes, restored silently, under precisely the exhaustion the gate watches for. Now an unavailable buffer is itself a fault. Cleanup moved to a `trap`; the previous bare `rm` did not survive a tick killed for overrunning its launchd window, so the earlier claim of cleanup "on every path" was false.

**`b522ab6` — a test arm that passed against the bug.** The piped-stdout arm returned 240 of 240 on the reverted implementation while its own comment called it "the arm that fails." Deleted. Also closed real missing coverage: no fixture had a space in a *directory* name, so `awk -F'\t'` — load-bearing, because this vault has folders like `Awesome Motive` — survived deletion with the suite green. The fd sweep is down to the one limit that discriminates.

**`053a627` — three false comments**, including one that named `last_scan` as driving the owner-election window. It has no such consumer, and that claim was the gate's stated justification.

The panel's remaining findings are filed as #51 (the fault predicate keys on stderr, so it latches on noise and misses quiet faults), #52 (cold-start banner silence), #53 (quarantine stderr outside the capture), #54 (fault-buffer dedup before truncation), #55 (threshold coercion), #56 (vault-root absolute-path leak), and #57 (end-to-end tick test under a low fd limit).

## Testing Procedure

1. Check out this branch and run `bash tests/run-all.sh` — expect `ALL PASS` (bash 3.2.57 verified).
2. Revert `scripts/lib/vault-scan.sh` to master's version and run `bash tests/vault-scan.sh`. Expect `FAIL: ulimit 64: got 102 clusters, want 240 (scan truncated silently)`. The test builds its own 120-directory fixture (3 files each, 2 shared tokens = exactly 240 clusters) and asserts that count at `ulimit -n 64` — the only limit that discriminates; 256 and the default both passed against the bug and were dropped in `b522ab6`, along with a piped-vs-redirected arm that also passed against it.
3. Delete the `if [ -n "$SCAN_FAULT" ]` block from `scripts/vaultkeeper-tick.sh` and run `bash tests/vaultkeeper-tick.sh`. Expect `FAIL: a faulted scan was not named incomplete`. That arm shadows one scanner with a version that writes to stderr and returns 0 — the exact shape every scanner has today — then asserts the tick stays 0, quotes the scanner, says INCOMPLETE, never says complete, and leaves `last_scan` absent.
4. To see the original symptom on real data: `bash -c "ulimit -n 128; . scripts/lib/vault-scan.sh; scan_clusters <your-vault> 3 | wc -l"` on master vs this branch.

## Additional Notes / HelpScout Tickets

Measuring against the live vault is useless — Syncthing and the keeper itself mutate it mid-run, and three consecutive runs gave 211, 336 and 503. Every number above comes from a frozen copy or a synthetic fixture. I burned several measurement rounds on that before noticing, and one intermediate claim ("production is losing lines") was wrong as a result; the frozen-fixture numbers are the ones that hold.

Related: #43 (temp-swap invariant, the Pending-churn half), #44, #45, #46.

---

Two corrections to the commits themselves, for the record. `048359e`'s message says it "retires the `${SCAN_ERR:-/dev/stderr}` fallback" — it does not; the fallback is still there and is now only reachable once a fault is already recorded, so it is inert rather than removed. And the gate split left `keeper_quarantine_conflicts` on the wrong side: its `QUARANTINE` row is a one-shot side effect of an irreversible `mv`, so a faulted tick drops that row from `Pending.md` permanently. The conflict file is safely quarantined either way. Filed as #58 rather than bodged in, because deferring quarantine past the gate would empty the digest's quarantine section.

### DIFF
--- a/scripts/lib/surfacing.sh
+++ b/scripts/lib/surfacing.sh
@@ -3,13 +3,19 @@
 # append only genuinely-new candidates to Pending.md (transition-gated against
 # the prior-scan snapshot). Requires note-hash.sh + keeper-state.sh.
 
+# surfacing_digest <vault> [status]
+# `status` is stamped as `scan_status:` when non-empty. The caller passes INCOMPLETE
+# when the scanners faulted: this file is the human-facing artifact and it was
+# claiming a fresh full scan — timestamp AND section counts — on ticks the keeper
+# itself refused to record as complete.
 surfacing_digest() {
-  local vault="$1" lines tmp target="$1/Librarian.md" kind label
+  local vault="$1" status="${2:-}" lines tmp target="$1/Librarian.md" kind label
   lines="$(cat)"
   tmp="$(mktemp "$vault/.Librarian-XXXXXX")" || return 1
   {
     printf '%s\n' '<!-- MACHINE-OWNED: regenerated each vaultkeeper scan. Edits do not persist. -->'
     printf '# Librarian\n\nlast_scan: %s\n' "$(now_epoch)"
+    [ -n "$status" ] && printf 'scan_status: %s\n' "$status"
     for kind in GAP UNFILED ASK CLUSTER QUARANTINE; do
       if [ "$kind" = "GAP" ]; then label="Frontmatter gaps"
       elif [ "$kind" = "UNFILED" ]; then label="Unfiled (Inbox)"
--- a/scripts/lib/vault-scan.sh
+++ b/scripts/lib/vault-scan.sh
@@ -72,10 +72,12 @@ scan_open_asks() {
 # One find over the whole vault, grouped by (dir, token) at the end. The previous
 # shape walked directories and opened a nested process substitution per directory
 # — two live fds per iteration — which exhausted the fd table under launchd and
-# truncated the candidate set while still returning 0. It also lost 58% of its
-# output whenever stdout was a plain pipe, with nothing on stderr. Emitting flat
-# `dir<TAB>token` pairs through a single pipeline has one process substitution for
-# the entire scan, so neither failure has anywhere to happen.
+# truncated the candidate set while still returning 0. How much it dropped tracked
+# whatever fd headroom happened to exist at runtime: the same frozen 569-directory
+# vault came back with 503 clusters on an idle host and 211 under load, identical
+# through a pipe and through a file redirect, and with nothing on stderr in either
+# case. Emitting flat `dir<TAB>token` pairs through a single pipeline leaves one
+# process substitution for the whole scan, so there is no headroom to run out of.
 scan_clusters() {
   local vault="$1" threshold="$2" f dir base slug rel
   while IFS= read -r f; do
--- a/scripts/lib/vault-scan.sh
+++ b/scripts/lib/vault-scan.sh
@@ -69,31 +69,29 @@ scan_open_asks() {
 # >= threshold distinct files. No associative arrays (bash 3.2 on the macOS
 # keeper host) — count via sort|uniq -c. Tokens are unique-per-file (sort -u),
 # so uniq -c across files = number of distinct files sharing the token.
+# One find over the whole vault, grouped by (dir, token) at the end. The previous
+# shape walked directories and opened a nested process substitution per directory
+# — two live fds per iteration — which exhausted the fd table under launchd and
+# truncated the candidate set while still returning 0. It also lost 58% of its
+# output whenever stdout was a plain pipe, with nothing on stderr. Emitting flat
+# `dir<TAB>token` pairs through a single pipeline has one process substitution for
+# the entire scan, so neither failure has anywhere to happen.
 scan_clusters() {
-  local vault="$1" threshold="$2" dir base f tok cnt
-  while IFS= read -r dir; do
-    while read -r cnt tok; do
-      [ -z "$cnt" ] && continue
-      [ "$cnt" -ge "$threshold" ] \
-        && printf 'CLUSTER\t%s\t%s\t%s\n' "${dir#"$vault"/}" "$tok" "$cnt"
-    done < <(
-      while IFS= read -r f; do
-        base="$(basename "$f" .md)"
-        # Spaces are folded to the same separator before tokenizing: tokenize_slug
-        # splits on `-` only, and this vault has filenames with spaces. `sort -u`
-        # keeps the count "distinct files containing the token", not total
-        # occurrences.
-        printf '%s\n' "$base" | tr ' ' '-' | while IFS= read -r slug; do
-          tokenize_slug "$slug"
-        done | sort -u
-      done < <(find "$dir" -maxdepth 1 -type f -name '*.md') \
-        | sort | uniq -c
-    )
-  done < <(find "$vault" -type d \
-            ! -path '*/.obsidian' ! -path '*/.obsidian/*' \
-            ! -path '*/.trash'    ! -path '*/.trash/*' \
-            ! -path '*/.git'      ! -path '*/.git/*' \
-            ! -path '*/.vaultkeeper-quarantine' ! -path '*/.vaultkeeper-quarantine/*' \
-            ! -path '*/.vaultkeeper'            ! -path '*/.vaultkeeper/*')
+  local vault="$1" threshold="$2" f dir base slug rel
+  while IFS= read -r f; do
+    dir="${f%/*}"
+    base="${f##*/}"; base="${base%.md}"
+    # Spaces are folded to the same separator before tokenizing: tokenize_slug
+    # splits on `-` only, and this vault has filenames with spaces. `sort -u`
+    # keeps the count "distinct files containing the token", not total occurrences.
+    slug="${base// /-}"
+    rel="${dir#"$vault"/}"
+    tokenize_slug "$slug" | sort -u | while IFS= read -r tok; do
+      [ -n "$tok" ] && printf '%s\t%s\n' "$rel" "$tok"
+    done
+  done < <(_scan_find_md "$vault") \
+  | sort | uniq -c \
+  | sed 's/^ *\([0-9][0-9]*\) /\1'"$(printf '\t')"'/' \
+  | awk -F'\t' -v t="$threshold" '$1+0 >= t+0 { printf "CLUSTER\t%s\t%s\t%s\n", $2, $3, $1 }'
   return 0
 }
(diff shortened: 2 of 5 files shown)


nhangen (commented):
**Self-review, so the event is `COMMENT`.** Panel ran 6 of 8 agents — the Skeptic and Domain Expert both died on API/session limits. Treat items 1–4 as merge blockers.

**The fd fix holds.** No agent could break it. Under host load where the old code returns 211 of 503 clusters, the rewrite returns 503 three times on both output targets; one reviewer measured it 2.4× faster. Byte-for-byte line-set equality with the old implementation on 2720 real notes.

### 1. `vaultkeeper-tick.sh:64` — the gate is below the writes it guards

`base_view_write` (`:61`), `surfacing_digest` (`:62`) and `surfacing_pending_transition` (`:63`) all run first. So a tick that declares itself INCOMPLETE has already overwritten the snapshot with the truncated candidate set, and the next healthy tick re-appends the missing rows as new. Three agents measured the duplicate growth independently — 3 dups over 4 cycles, 12 over 4 cycles. That flap is *new*, introduced by this commit: the pre-PR truncation was stable, so it never oscillated.

The fix isn't a plain move — a plain move leaves a cold-start host with a permanent fault holding no `Librarian.md` at all, forever. Split it: keep `base_view_write` and `surfacing_digest` above the gate, move only `surfacing_pending_transition` below, and mark the digest `scan_status: INCOMPLETE`.

### 2. `:45` — `mktemp` failure silently disables the entire gate

`SCAN_ERR=""` → stderr falls back to `/dev/stderr` → `SCAN_FAULT` stays empty → `keeper_record_scan` runs → `scan complete`. That's the 737-tick bug restored, and `mktemp` is among the first things to fail under the fd exhaustion this gate exists to catch. Fail closed.

### 3. The mechanism claim in the PR body is wrong, and so was every theory the panel proposed

I claimed the loss was pipe-specific and independent of the fd limit. Four reviewers agreed, one claimed the reverse direction, the Shipper said fd-gated, I concluded depth-driven. All wrong. The same frozen vault with the same old code returns **503 clusters on a quiet host and 211 under load, with pipe == file at 211**. It is ambient fd headroom.

The true claim is simpler: *old `scan_clusters` silently returns an unpredictable fraction of its clusters depending on available fds — 211 to 503 of 503 for one vault — with zero stderr and rc=0.*

### 4. `tests/vault-scan.sh:199` — the piped arm passes against the un-fixed code

240 of 240 on reverted `vault-scan.sh`. It claims in a comment to be "the arm that fails." Growing the fixture doesn't help — there is no target asymmetry to catch. Delete the arm and the claim; the `ulimit 64` squeeze is the only deterministic pin and it already works (102 vs 240).

### Also fixing

- The comment claiming an unrecorded `last_scan` makes the banner report stale — it makes it *silent*, and the repo's own `tests/keeper-state.sh:37` asserts exactly that.
- The invented "owner-election window" consumer. `last_scan` has no such consumer; election uses claim-file mtimes. That claim was the gate's stated justification, in three places.
- `surfacing_digest` stamping a fresh `last_scan` **and** `## Open [ask] items (0)` into the synced `Librarian.md` on an INCOMPLETE tick — so the counts lie, not just the timestamp.
- The missing space-bearing folder fixture, without which `awk -F'\t'` survives deletion while production has `Awesome Motive`.

### Filing separately

The any-stderr predicate latching forever on one unreadable directory; a fault with *no* stderr (signal death, quiet non-zero) being invisible to the gate entirely; `staleness_banner` going silent instead of warning on a cold start; `keeper_quarantine_conflicts` sitting outside the capture; fault text truncated after being repeated once per scanner; `awk` coercing a non-numeric threshold to 0; "byte-identical" being wrong on ordering; no `trap` on the temp file; and the vault-root absolute-path leak into `Librarian.md`.

nhangen:
The fd fix and the piped-stdout arm both look right, and the two-failure-mode test is the part that makes it stick. Two things about the gate, though — as written it doesn't accomplish what the comment says it does.

**`last_scan` has two writers, and only one of them is gated.** `surfacing_digest` writes `last_scan: <now>` into `Librarian.md` (`scripts/lib/surfacing.sh:12`), and in `vaultkeeper-tick.sh` it runs *before* the `SCAN_FAULT` check. So a faulted scan still stamps a fresh timestamp into the digest that agents and I actually read — the exact lie the gate is there to prevent. The cache-side `keeper_record_scan` is withheld correctly; the vault-side one isn't. Moving both surfacing calls below the gate, or passing the fault down into the digest so it renders `last_scan: INCOMPLETE`, would close it.

**The banner doesn't fire when `last_scan` is absent.** `staleness_banner` bails on `[ -z "$last" ] && return 0` (`keeper-state.sh:48-49`). The comment says leaving it unrecorded "makes the banner report stale, which is the true state" — that only holds on a host that already has a good scan on disk. On a cold start, or after a cache wipe, if the early scans fault there's no `last_scan` at all and `/obsidian:ask` prints nothing. Never-scanned and just-scanned look identical to the consumer. Probably wants a distinct "never completed a scan" line rather than a silent return.

Smaller one: `surfacing_pending_transition` is also above the gate, so a truncated candidate set gets written as the prior-scan snapshot. Next complete tick then sees the entries the truncated scan missed as brand-new and appends them to `Pending.md`. Same fix as the first point — move it below.

Issue #58, opened 2026-07-29
A faulted tick drops the QUARANTINE row from Pending.md permanently


nhangen:
Introduced by PR #49 and found by its fix auditor. Not shipped-broken elsewhere — this is the one write site the gate split left on the wrong side.

`keeper_quarantine_conflicts` (`scripts/vaultkeeper-tick.sh:38`, defined in `keeper-lease.sh:52`) `mv`s a Syncthing conflict file into `.vaultkeeper-quarantine/` and emits its `QUARANTINE\t<path>` row exactly once — the row is a side effect of the move, not a re-derivable observation.

#49 moved `surfacing_pending_transition` below the scan-fault gate so a truncated candidate set can no longer rewrite the snapshot. Correct for the four scanners, whose rows are re-derived every tick. Wrong for this one: if the tick faults, the move has already happened, the gate blocks the transition, and no later tick can re-emit the row because `find` no longer sees the file (it is excluded via `! -path "*/.vaultkeeper-quarantine/*"`).

Consequence is bounded — the conflict file is safely quarantined either way, so this costs the `- [ ] QUARANTINE: …` line in `Pending.md`, not the data. It needs both a real sync conflict and a faulted tick on the same run.

The fix is not a one-line move, which is why it is filed rather than bodged into #49: quarantine runs before the digest so its rows appear in `Librarian.md`, so deferring it past the gate would empty the digest section, while letting its rows through the gate alone would reopen the snapshot-rewrite path. Options worth weighing: give the transition a subset that only ever appends (never overwrites the snapshot) for irreversible-side-effect rows; or make quarantine idempotent by recording quarantined paths in state so a later tick can re-emit.

Issue #54, opened 2026-07-29
Fault buffer is truncated to 300 chars after the same message repeats once per scanner


nhangen:
From the #49 review panel. `_scan_find_md` is called by three scanners, so one unreadable directory produces three identical `find: … Permission denied` lines in the buffer. `${SCAN_FAULT:0:300}` then truncates *after* the repetition, mid-word, so one repeated benign error can consume the whole budget and hide a genuinely serious fault later in the buffer — degrading the diagnostic the gate exists to deliver.

Fix: `sort -u` (or `uniq`) the buffer before truncating.

Issue #53, opened 2026-07-29
keeper_quarantine_conflicts runs outside the scan-fault capture


nhangen:
From the #49 review panel. `scripts/vaultkeeper-tick.sh` computes `CONFLICTS="$(keeper_quarantine_conflicts "$VAULT" || true)"` before the `2>"$SCAN_ERR"` capture opens, and injects its `QUARANTINE` rows into `CAND`.

So a failed quarantine move — `vault-scan.sh` prints `failed to quarantine %s` to stderr and `continue`s — produces a short QUARANTINE list, the stderr escapes the gate, and the `|| true` discards the status. The tick then records the scan as complete. Same invariant as #49, one function over.

Fix: move the assignment inside the braced group, or redirect it into the same buffer.

Issue #51, opened 2026-07-29
Scan-fault gate keys on stderr, so it over-triggers on noise and misses faults that are quiet


nhangen:
From the #49 review panel. The gate added in that PR treats any byte on the scanners' stderr as a fault. Two failure modes fall out of using a diagnostic channel as a control signal.

**Over-triggers, and latches.** One `chmod 000` directory makes `find` print `Permission denied`, so every tick from then on prints INCOMPLETE and never records `last_scan`. Four agents reproduced it across consecutive ticks with no recovery path. The scan was otherwise complete — every readable note was scanned. Not exotic on the target hosts: a vault under iCloud or with a stale network mount yields `Operation not permitted` from macOS TCC on paths the tick can never fix.

**Under-triggers.** A scanner killed by a signal, or returning non-zero quietly, produces no stderr and is invisible to the gate. `scripts/lib/vault-scan.sh` scanners all end in a hardcoded `return 0`, so there is still no status channel — the gate catches the chatty half of the problem only. This was the thread a review agent was pulling when it died.

**Do not fix with a fatal-pattern allowlist.** Matching `Too many open files` / `cannot duplicate fd` fails *silent* when the list is too narrow, which is the bug class #49 exists to remove, hand-rebuilt. A noise *denylist* (`Permission denied`, `Operation not permitted`) fails loud instead, which is the safe direction if a pattern is used at all.

The structural fix is to give the scanners a real completeness signal — e.g. `_scan_find_md` is now the single choke point for all four, so a truncation marker set there would be inherited by every caller instead of bolted onto one call site.

PR #59, opened 2026-07-29
Make a never-completed scan visible to the staleness banner

### LINKED ISSUE #52: staleness_banner is silent when last_scan is absent, so a never-scanned host looks healthy
From the #49 review panel. `staleness_banner` (`scripts/lib/keeper-state.sh`) opens with `last="$(keeper_last_scan "$vault")"; [ -z "$last" ] && return 0`, so an absent `last_scan` prints nothing — indistinguishable from a fresh successful scan. `tests/keeper-state.sh` enshrines it (`|| fail "no banner when last_scan absent"`).

Measured: absent → `[]`, stale → `[⚠ index last maintained 99999s ago]`, fresh → `[]`.

This is the other half of the gate problem. #49 withholds `last_scan` when the scan faulted, which works on a host that has already recorded one — the banner fires when the old value ages past 2×interval. On a **fresh install whose first scan faults** the value is never written, so `/obsidian:ask` reports nothing wrong, permanently. Combined with the latching predicate in the sibling issue, such a host is invisibly degraded forever, which is worse than the over-reporting it replaced.

Fix: treat a missing `last_scan` as maximally stale, or emit a distinct cold-start warning when `last_scan` is absent but the cache dir exists. Note the existing test asserts the current behavior and will need updating with it.

nhangen:
Closes #52

## Description

Follow-up to #49. That PR made the tick withhold `last_scan` when the scanners fault, so `/obsidian:ask` would report the index as stale rather than claim a scan it didn't finish. The withholding was correct; the consumer couldn't hear it.

`staleness_banner` returned early on an absent `last_scan` (`keeper-state.sh`, `[ -z "$last" ] && return 0`). On a host whose scans fault from the *first* tick there is no previous `last_scan` to age out, so the banner printed nothing — indistinguishable from a vault scanned a second ago. A fault latched from tick one was silent.

An absent `last_scan` is two different states, and only one of them earns silence:

- **nothing has tried yet** — fresh install, or a host that defers to the elected owner and never scans. Silence is right.
- **every attempt faulted** — exactly what the #49 gate withholds `last_scan` to signal. This is the loud case.

So the tick now records an attempt (`keeper_record_attempt`) below the ownership gate and before scanning, and the banner uses it to tell the two apart. With an attempt and no completed scan it prints `⚠ index has never completed a scan (last attempt Ns ago)`. No aging grace on that branch on purpose — one faulted tick already means the index was never fully built, and the in-flight window between recording the attempt and recording the scan is the couple of seconds a healthy tick takes.

Recording below the ownership gate matters: doing it above would have every non-owner host claim an attempt it never makes, and then permanently report "never completed a scan" for a vault it isn't responsible for scanning.

Also updates the gate comment in `vaultkeeper-tick.sh`, which documented this gap as unfiled.

## Testing Procedure

1. Check out this branch.
2. `bash tests/keeper-state.sh` and `bash tests/vaultkeeper-tick.sh` — both pass.
3. Confirm the assertions bind, one mutation at a time:
   - Restore the early return: change `if [ -z "$last" ]; then` to `if [ -z "$last" ]; then return 0;` in `scripts/lib/keeper-state.sh`. `keeper-state.sh` fails with `attempted-but-never-completed host must warn, got silence`; `vaultkeeper-tick.sh` fails with `a host whose only scans faulted still looks healthy to the banner`.
   - Delete the `keeper_record_attempt "$VAULT"` line from `scripts/vaultkeeper-tick.sh`. `vaultkeeper-tick.sh` fails with `a faulted tick recorded no attempt, so the banner cannot tell it from a fresh install`.
4. Confirm the fresh-install case stays quiet — `tests/keeper-state.sh:37` (`no banner when last_scan absent`) is unchanged and still passes, as does the never-attempted assertion on the new fixture vault.
5. Full suite: `for t in tests/*.sh; do bash "$t" >/dev/null || echo "FAIL $t"; done` — 29/29, no output.

## Notes

Both #49 findings that touched `Librarian.md` and `Pending.md` are already fixed on master (`scan_status: INCOMPLETE` and moving `surfacing_pending_transition` below the gate). This PR is the third one, which #49 knowingly deferred to #52.

The digest still writes a fresh `last_scan:` on a faulted tick alongside `scan_status: INCOMPLETE`. Left as-is — the field means "when the scan ran", and the status line carries whether it finished.

Open and unrelated to this change: #58 (a faulted tick drops the QUARANTINE row permanently), #53, #51, #54.

### DIFF
--- a/scripts/vaultkeeper-tick.sh
+++ b/scripts/vaultkeeper-tick.sh
@@ -35,6 +35,13 @@ if ! keeper_is_owner "$LEASE" "$HOST" "$PRIORITY" "$MAXAGE"; then
   exit 0
 fi
 
+# Record the attempt below the ownership gate, so only the host that actually scans
+# claims one. staleness_banner needs it to read a missing last_scan correctly: with
+# no attempt it means nothing has tried yet (fresh install, or a host deferring to
+# the owner) and silence is right; with an attempt it means every try faulted, which
+# is the state the gate below withholds last_scan to signal.
+keeper_record_attempt "$VAULT"
+
 CONFLICTS="$(keeper_quarantine_conflicts "$VAULT" || true)"
 # Capture the scanners' stderr instead of letting it fly past. Every scanner
 # returns 0 unconditionally, so a truncated scan was indistinguishable from a
@@ -87,9 +94,10 @@ if [ -n "$SCAN_FAULT" ]; then
   # this check sat below it, in a file the user hand-edits and Syncthing replicates.
   # keeper_record_scan is withheld for the ordinary reason: last_scan's consumer is
   # staleness_banner (scripts/ask-staleness.sh), and a partial scan recorded as
-  # complete tells it the vault was fully examined. Note the banner only fires once a
-  # PREVIOUS last_scan ages out — a host that has never recorded one stays silent, so
-  # a latched fault here is invisible rather than loud. That gap is filed separately.
+  # complete tells it the vault was fully examined. Withholding it now actually
+  # reaches that consumer: the attempt recorded above the scan lets the banner tell a
+  # host whose every tick faulted from one that has never tried, so a fault latched
+  # from the first tick reads as "never completed a scan" instead of as silence.
   printf 'vaultkeeper: scan INCOMPLETE on %s — snapshot and Pending.md left untouched; scanners said: %s\n' \
     "$HOST" "$SCAN_FAULT" >&2
   exit 0
--- a/scripts/lib/keeper-state.sh
+++ b/scripts/lib/keeper-state.sh
@@ -43,11 +43,30 @@ keeper_last_scan_file() { printf '%s/last_scan\n' "$(keeper_cache_dir "$1")"; }
 keeper_record_scan()    { local f; f="$(keeper_last_scan_file "$1")"; mkdir -p "$(dirname "$f")"; now_epoch > "$f"; }
 keeper_last_scan()      { local f; f="$(keeper_last_scan_file "$1")"; [ -f "$f" ] && cat "$f" || true; }
 
+# An attempt is recorded by the elected owner before it scans, whether or not that
+# scan goes on to complete. It is what separates "no last_scan because nothing has
+# ever tried" (fresh install, or a host that defers to the owner — silence is right)
+# from "no last_scan because every attempt faulted" (the scan-fault gate withholding
+# it, which is exactly the state worth shouting about).
+keeper_last_attempt_file() { printf '%s/last_attempt\n' "$(keeper_cache_dir "$1")"; }
+keeper_record_attempt()    { local f; f="$(keeper_last_attempt_file "$1")"; mkdir -p "$(dirname "$f")"; now_epoch > "$f"; }
+keeper_last_attempt()      { local f; f="$(keeper_last_attempt_file "$1")"; [ -f "$f" ] && cat "$f" || true; }
+
 staleness_banner() {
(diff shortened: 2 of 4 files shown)


scripts/vaultkeeper-tick.sh
```diff
@@ -35,6 +35,13 @@ if ! keeper_is_owner "$LEASE" "$HOST" "$PRIORITY" "$MAXAGE"; then
   exit 0
 fi
 
+# Record the attempt below the ownership gate, so only the host that actually scans
+# claims one. staleness_banner needs it to read a missing last_scan correctly: with
+# no attempt it means nothing has tried yet (fresh install, or a host deferring to
+# the owner) and silence is right; with an attempt it means every try faulted, which
+# is the state the gate below withholds last_scan to signal.
+keeper_record_attempt "$VAULT"
```
nhangen, line 43: This placement is the PR body's central claim, and nothing holds it. Move this line above the ownership gate and all 29 suites still pass — but a deferring host then prints the never-completed banner forever, on a vault the owner scanned fine. The existing non-owner case at tests/vaultkeeper-tick.sh:38-41 can't catch it because `$V` already has a `last_scan` from the owner tick at :27, so the banner takes the aging branch instead. Needs a fixture on a vault the non-owner has never scanned, asserting no attempt file is written and the banner stays silent.

nhangen (commented):
Panel ran three reviewers plus an auditor that verified every cited `file:line` against the branch. No hallucinations. Two findings worth acting on before merge, and the interesting one is a conflict between two of them.

Both reviewers landed on `keeper_record_attempt`'s placement and wanted opposite things — one said the deferring host should speak up when the owner is faulting, the other said the suite doesn't pin that it stays quiet. The probe settles it: move that call above the ownership gate and a deferring host prints `⚠ index has never completed a scan` on a vault the owner just scanned successfully. Permanent false alarm, no recovery path, and a banner that always fires carries no information. So the placement as committed is right — and the whole suite stays green when you break it, which is the actual problem.

The cross-host blindness is real, though. A deferring host genuinely can't see an owner that has faulted since install, and on a two-host setup the deferring host is usually where you're asking. Master behaves the same way and it isn't what #52 asked for, so it's a follow-up rather than a blocker — and the signal already exists in the vault (`Librarian.md` carries `scan_status: INCOMPLETE`, the lease names the owner), so it doesn't need new state.

Smaller stuff this PR made wrong: the module header at `scripts/lib/keeper-state.sh:2-3` names every state artifact except the one just added, and `obsidian.local.md.example` still says the banner keys off `2x keeper_interval_secs` — the new branch never reads the interval.

Filed rather than fixed here, since the pre-existing `last_scan` path has it identically: both epoch readers treat an empty file as absent and abort on a non-numeric one. A zero-byte `last_attempt` (truncated ENOSPC write) reads as "never attempted" and goes silent; a non-numeric one exits 1 with no stdout, which every consumer reads as healthy. Wants one validated-read helper across both paths, not a patch on the new one.

Verdict downgraded to a comment rather than changes-requested — self-authored PR, GitHub won't take either verdict from its own author.

nhangen:
Superseded by #97. The #52 fix landed on master while this panel was running, so this branch went 20 commits behind and CONFLICTING — and master's fix keys the never-completed test off the cache dir, which `keeper_state_init` creates above the ownership gate, so a deferring host warns forever about a vault the owner is scanning fine. #97 is this work rebuilt on current master: same attempt-file mechanism, master's unreadable guard kept and extended, plus the deferring-host fixture that fails anyone who reintroduces the false alarm.

PR #102, opened 2026-08-21
save-conversation: add fast path for routine execution saves

nhangen:
## Summary

Adds a guarded fast-path section to the top of `save-conversation/SKILL.md` so routine execution/project saves stop paying ~7k tokens reading contingency machinery (cross-domain tiebreakers, substrate filing, MOC promotion, chapter segmentation) that doesn't apply.

## What changes

- New ~55-line **Fast Path** section immediately after the skill intro
- Fail-closed guards: topic hint, 2+ taxonomy matches, dedup ≥ threshold, research/planning/reflection sessions, `#bookmark` markers, Inbox routing — any hit drops to the unchanged full procedure
- Fast path calls the same validators (`allowlist_validate`, `dedup_same_day`, `kspayload_validate`) and the same keeper CLI insert with `resolved: true` — no second writer, no skipped gate; it is a reading-order optimization only
- Inlines the payload format + keeper CLI invocation so `keeper-save/SKILL.md` need not be read on the insert path
- Version bump 1.2.0 → 1.3.0

## Testing

- `tests/run-all.sh` in a worktree: all suites pass except `commit-capture-head-gate.sh` and `vaultkeeper-tick.sh`, both confirmed failing on pristine `origin/master` (pre-existing, unrelated)
- Touched-area suites green: `sole-brain-save-conversation`, `keeper-save-payload`, `dedup-scan`, `resolve-config`, `moc-promote`

Environment: local test fixtures via `tests/run-all.sh`.

### DIFF
--- a/skills/save-conversation/SKILL.md
+++ b/skills/save-conversation/SKILL.md
@@ -1,7 +1,7 @@
 ---
 name: save-conversation
 description: Exports the current Claude conversation to the Obsidian vault. Triggers on phrases like "save this to Obsidian", "export to obsidian", "document this session", "save our conversation", "write this up", "put this in obsidian". Formats the conversation as structured markdown, determines the correct project folder from context, and saves with a timestamped filename. Optionally opens the note in the Obsidian GUI.
-version: 1.2.0
+version: 1.3.0
 ---
 
 # Save Conversation to Obsidian
@@ -16,6 +16,60 @@ directly. The keeper is the sole writer; this skill resolves the target and
 passes a `resolved: true` payload so the keeper writes it as-is without
 re-routing or re-deduping. The one exception is MOC-promotion *file moves*, which
 remain native (structural reorg is the vault-organizer's domain, not a note-write).
+
+## Fast Path (routine execution/project saves)
+
+Use this path when **all** guards hold; otherwise drop to the full procedure below.
+The fast path calls the same validators and the same keeper CLI as the full
+procedure — nothing is skipped, only the reading of the detail sections.
+
+**Guards (any hit → full procedure):**
+- A topic hint was provided
+- 2+ taxonomy domains matched the keyword scan (needs the Cross-Domain Tiebreaker)
+- Same-day dedup score ≥ threshold (needs append vs new prompt)
+- Session was research/planning/reflection, or tested an active research claim
+  (needs intent scoring care / substrate filing)
+- `#bookmark` markers or >30-min gaps present (Chapter Segmentation)
+- Target folder would be `Inbox/` (ambiguous — needs confirmation)
+
+**Steps:**
+```bash
+CONFIG="$(bash "${CLAUDE_PLUGIN_ROOT}/scripts/lib/resolve-config.sh")"
+# route: single keyword match against ## Project Taxonomy in $CONFIG → <target-folder>
+. "${CLAUDE_PLUGIN_ROOT}/scripts/lib/allowlist-validate.sh"
+allowlist_validate "<target-folder>" || exit 0   # refusal already printed
+. "${CLAUDE_PLUGIN_ROOT}/scripts/lib/dedup-scan.sh"
+dedup_same_day "<vault_path>/<target-folder>" "$(date +%Y-%m-%d)" "<slug>" "${dedup_jaccard_threshold:-0.4}"
+# no output above → proceed
+```
+
+Build the body from the frontmatter template in **Output Format** below
+(`session_intent: execution`, `capture_action: project_note`,
+`research_state_change: none`), then write through the keeper:
+
+```
+op: insert
+resolved: true
+target: <target-folder>/<YYYY-MM-DD-title>.md
+title: <YYYY-MM-DD-title>
+session_link_date: <today YYYY-MM-DD>
+---
+<frontmatter + body per Output Format>
+```
+
+Validate with `kspayload_validate <payload-file>` (source
+`${CLAUDE_PLUGIN_ROOT}/scripts/lib/keeper-save-payload.sh`), extract the body
+with `kspayload_body`, then run the keeper CLI directly:
+
+```bash
+bash "${CLAUDE_PLUGIN_ROOT}/scripts/keeper" insert --vault "$VAULT" \
+  --target "<target-folder>/<YYYY-MM-DD-title>.md" \
+  --body-file "<body-temp>" --title "<YYYY-MM-DD-title>" \
+  --session-link-date "<today YYYY-MM-DD>"
+```
+
+Relay the committed path. Then run the MOC-Promotion Prompt check (step 14) —
+that post-save step applies on the fast path too.
 
 ## Config
 
--- a/skills/save-conversation/SKILL.md
+++ b/skills/save-conversation/SKILL.md
@@ -21,25 +21,35 @@ remain native (structural reorg is the vault-organizer's domain, not a note-writ
 
 Use this path when **all** guards hold; otherwise drop to the full procedure below.
 The fast path calls the same validators and the same keeper CLI as the full
-procedure — nothing is skipped, only the reading of the detail sections.
+procedure — nothing is skipped, only the reading of the detail sections. Three
+full-procedure steps are intentionally not restated here because they cannot
+change a routine execution save: the VAULT.md conventions read (step 1), the
+staleness banner, and `auto_open` handling (step 13) are covered by the notes
+below; when in doubt, use the full procedure.
 
 **Guards (any hit → full procedure):**
 - A topic hint was provided
 - 2+ taxonomy domains matched the keyword scan (needs the Cross-Domain Tiebreaker)
+- No taxonomy domain matched (route per Routing Logic / confirm Inbox with the user)
 - Same-day dedup score ≥ threshold (needs append vs new prompt)
 - Session was research/planning/reflection, or tested an active research claim
-  (needs intent scoring care / substrate filing)
+  (state your one-line evidence for this call; if borderline, use the full procedure)
 - `#bookmark` markers or >30-min gaps present (Chapter Segmentation)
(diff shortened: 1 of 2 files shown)


skills/save-conversation/SKILL.md
```diff
@@ -17,6 +17,60 @@ passes a `resolved: true` payload so the keeper writes it as-is without
 re-routing or re-deduping. The one exception is MOC-promotion *file moves*, which
 remain native (structural reorg is the vault-organizer's domain, not a note-write).
 
+## Fast Path (routine execution/project saves)
+
+Use this path when **all** guards hold; otherwise drop to the full procedure below.
+The fast path calls the same validators and the same keeper CLI as the full
+procedure — nothing is skipped, only the reading of the detail sections.
```
nhangen: The parity claim overreaches: the fast path also drops step 1 (VAULT.md conventions read), which no guard covers. Either narrow the sentence or add a guard for VAULT.md.

skills/save-conversation/SKILL.md
```diff
@@ -17,6 +17,60 @@ passes a `resolved: true` payload so the keeper writes it as-is without
...
+
+**Guards (any hit → full procedure):**
+- A topic hint was provided
+- 2+ taxonomy domains matched the keyword scan (needs the Cross-Domain Tiebreaker)
+- Same-day dedup score ≥ threshold (needs append vs new prompt)
+- Session was research/planning/reflection, or tested an active research claim
```
nhangen, line 35: This guard is self-assessed with no recovery path: a research-flavored session misread as routine bakes `research_state_change: none` into frontmatter permanently. Requiring a stated evidence line here (mirroring the Output Format evidence bullet) closes it cheaply.

skills/save-conversation/SKILL.md
```diff
@@ -17,6 +17,60 @@ passes a `resolved: true` payload so the keeper writes it as-is without
...
+  (needs intent scoring care / substrate filing)
+- `#bookmark` markers or >30-min gaps present (Chapter Segmentation)
+- Target folder would be `Inbox/` (ambiguous — needs confirmation)
+
+**Steps:**
+```bash
+CONFIG="$(bash "${CLAUDE_PLUGIN_ROOT}/scripts/lib/resolve-config.sh")"
+# route: single keyword match against ## Project Taxonomy in $CONFIG → <target-folder>
```
nhangen, line 47: Guards cover 2+ matches and Inbox targets but not zero matches — a topic that hits no domain keyword has no stated destination on this path.

skills/save-conversation/SKILL.md
```diff
@@ -17,6 +17,60 @@ passes a `resolved: true` payload so the keeper writes it as-is without
...
+. "${CLAUDE_PLUGIN_ROOT}/scripts/lib/allowlist-validate.sh"
+allowlist_validate "<target-folder>" || exit 0   # refusal already printed
```
nhangen: Worth carrying over the full procedure's config-fault note: a refusal naming /obsidian:setup means the config is broken, not that the folder choice was wrong.

skills/save-conversation/SKILL.md
```diff
@@ -17,6 +17,60 @@ passes a `resolved: true` payload so the keeper writes it as-is without
...
+. "${CLAUDE_PLUGIN_ROOT}/scripts/lib/dedup-scan.sh"
+dedup_same_day "<vault_path>/<target-folder>" "$(date +%Y-%m-%d)" "<slug>" "${dedup_jaccard_threshold:-0.4}"
```
nhangen: `${dedup_jaccard_threshold:-0.4}` ignores the documented per-vault frontmatter override — nothing exports it. Pre-existing at line 327 and in vault-librarian too; fixing the plumbing properly belongs in resolve-config.sh, so tracking as a follow-up rather than patching one site here.

skills/save-conversation/SKILL.md
```diff
@@ -17,6 +17,60 @@ passes a `resolved: true` payload so the keeper writes it as-is without
...
+(`session_intent: execution`, `capture_action: project_note`,
+`research_state_change: none`), then write through the keeper:
+
+```
+op: insert
+resolved: true
+target: <target-folder>/<YYYY-MM-DD-title>.md
+title: <YYYY-MM-DD-title>
+session_link_date: <today YYYY-MM-DD>
+---
```
nhangen, line 66: Same-minute double save dies opaquely here: keeper insert hard-fails on an existing target and the fast path omits the `-2` collision suffix rule from the full procedure.

nhangen (commented):
Mini panel (3 specialists + auditor), all anchors verified against 5280210. Self-authored PR, so the APPROVE verdict GitHub won't let me cast is recorded here as a comment: nothing blocking, all findings are one-line prose fixes plus one test extension.

**Test coverage (the real one):** `tests/sole-brain-save-conversation.sh` never mentions the fast path — every symbol it greps exists in the untouched full-procedure sections, so reverting this PR entirely keeps all suites green (mutation-verified by two reviewers independently). The repo's own grep-prose test style extends naturally: fast-path presence, guard keywords, validate-before-insert ordering, and `vault_path` sourcing from `$CONFIG`.

The rest of the notes cluster around the fast path's "nothing is skipped" claim being slightly stronger than reality: VAULT.md conventions, the staleness banner, the collision fallback, and `auto_open` all drop without mention, `$VAULT` is used but never resolved, and two guard/gate contracts (zero-keyword routing, stop-on-validate-failure) go unstated. Each is a one-line prose addition. Two pre-existing issues (`dedup_jaccard_threshold` config plumbing across three call sites, keeper INDEX-link `|| true`) get follow-up tickets rather than scope creep here.

skills/save-conversation/SKILL.md
```diff
@@ -17,6 +17,60 @@ passes a `resolved: true` payload so the keeper writes it as-is without
...
+<frontmatter + body per Output Format>
+```
+
+Validate with `kspayload_validate <payload-file>` (source
```
nhangen, line 73: keeper-save states the contract this skips: on validation failure, surface stderr and stop. Without it the gate has no defined failure behavior, and keeper itself re-checks almost nothing content-bearing.

skills/save-conversation/SKILL.md
```diff
@@ -17,6 +17,60 @@ passes a `resolved: true` payload so the keeper writes it as-is without
...
+`${CLAUDE_PLUGIN_ROOT}/scripts/lib/keeper-save-payload.sh`), extract the body
+with `kspayload_body`, then run the keeper CLI directly:
+
+```bash
+bash "${CLAUDE_PLUGIN_ROOT}/scripts/keeper" insert --vault "$VAULT" \
```
nhangen, line 80: `$VAULT` is never resolved on this path — the steps set only `CONFIG`. keeper's resolve-config fallback masks it, but a literal reading hits an unset variable; one line reading `vault_path` from `$CONFIG` fixes it.

skills/save-conversation/SKILL.md
```diff
@@ -17,6 +17,60 @@ passes a `resolved: true` payload so the keeper writes it as-is without
...
+  --target "<target-folder>/<YYYY-MM-DD-title>.md" \
+  --body-file "<body-temp>" --title "<YYYY-MM-DD-title>" \
+  --session-link-date "<today YYYY-MM-DD>"
+```
+
+Relay the committed path. Then run the MOC-Promotion Prompt check (step 14) —
```
nhangen: Step 14 is carried forward explicitly but step 13 (`auto_open` GUI open) drops unmentioned — inconsistent for anyone running with `auto_open: true`.

skills/save-conversation/SKILL.md
nhangen: fixed in 4171c25 — parity claim now names what is intentionally dropped

skills/save-conversation/SKILL.md
nhangen: fixed in 4171c25 — guard now requires a stated evidence line

skills/save-conversation/SKILL.md
nhangen: fixed in 4171c25 — zero-keyword-match guard added

skills/save-conversation/SKILL.md
nhangen: fixed in 4171c25 — config-fault note carried over

skills/save-conversation/SKILL.md
nhangen: tracked #103 (pre-existing across three call sites; fix belongs in resolve-config.sh)

skills/save-conversation/SKILL.md
nhangen: fixed in 4171c25 — collision suffix rule documented on the fast path

skills/save-conversation/SKILL.md
nhangen: fixed in 4171c25 — stop-on-failure stated; gate ordering asserted by tests (5e0b1e7)

skills/save-conversation/SKILL.md
nhangen: fixed in 4171c25 — VAULT assigned from $CONFIG vault_path; assertion mutation-hardened in 5e0b1e7

skills/save-conversation/SKILL.md
nhangen: fixed in 4171c25 — auto_open honored after relaying the path

nhangen:
Dispositions for the panel findings that weren't anchored inline: **test coverage** (the HIGH finding) — fixed in 5e0b1e7; `tests/sole-brain-save-conversation.sh` now asserts fast-path presence, guard keywords, validate-before-insert ordering, and $CONFIG sourcing of VAULT/threshold as mutation-fatal assignments. Pre-existing keeper INDEX-link `|| true` swallow: tracked #104.

Issue #104, opened 2026-08-21
keeper insert swallows INDEX-link failure via || true


nhangen:
Found by the PR #102 panel. `scripts/keeper:168` runs `vault_index_apply … >/dev/null || true`: if INDEX linking fails the note is still written and the CLI exits 0 printing the committed path, so the caller relays success while the link never existed. Related partial-write window: the note is written before INDEX/session-link attempts, so a mid-insert failure can leave an unlinked note.

Fix: check vault_index_apply's rc and print a warning line on link failure (exit status can stay 0 if the note itself committed).

Context: https://github.com/nhangen/claude-obsidian-plugin/pull/102

nhangen:
Fixed by #105, merged as `00781d9`. The PR title listed the issue numbers but the body had no `Closes` line, so GitHub did not close them automatically.

Issue #103, opened 2026-08-21
dedup_jaccard_threshold config override is never plumbed into dedup_same_day


nhangen:
Found by the PR #102 panel. The per-vault `dedup_jaccard_threshold` frontmatter override (documented in `skills/save-conversation/SKILL.md:359`, written by setup) is never parsed into the environment or passed as an argument — every call site falls back to the 0.4 library default:

- `skills/save-conversation/SKILL.md` fast path + full procedure (`dedup_same_day ... ${dedup_jaccard_threshold:-0.4}`)
- `agents/vault-librarian.md:96` (passes no threshold arg)

A vault setting `0.0` (always prompt) silently runs at 0.4, re-enabling the twin-note failure the check exists to prevent. Proper fix belongs in `resolve-config.sh` (export it) so all three sites inherit it.

Context: https://github.com/nhangen/claude-obsidian-plugin/pull/102

nhangen:
Fixed by #105, merged as `00781d9`. The PR title listed the issue numbers but the body had no `Closes` line, so GitHub did not close them automatically.

Issue #101, opened 2026-08-13
keeper append --skip-if-hash can never match the heading commit-capture writes


nhangen:
`scripts/keeper`'s dedup gate requires the sha at the **end** of the heading:

```bash
grep -qE "^#{1,6}[[:space:]].*[[:space:]]${skip_hash}[[:space:]]*\$" "$full"
```

Every commit capture writes it at the **start** — `## <sha> — <message>` — so the gate never fires. Verified against a real note:

```
$ grep -qE "^#{1,6}[[:space:]].*[[:space:]]7ac9556[[:space:]]*$" \
    "$VAULT/Projects/Development/nhangen/llm-tools/2026-08-12.md" && echo MATCH || echo "NO MATCH"
NO MATCH
```

That note has `## 7ac9556 — repair the two guards #353 broke (llm-tools PR #357)` at line 739, so the section it is supposed to find is right there.

## Why it matters

`--skip-if-hash` is the *only* thing standing between a commit and a double capture, and the `obsidian-commit-capture` rule mandates passing it on every append for exactly that reason. It exits 0 either way, so a caller doing everything right gets no signal. The two cases it was added for (#62) are both still open: a re-run of the same capture, and a Syncthing conflict-copy re-capture on a second host.

It also fails in the safe direction, which is why it went unnoticed — a missed skip duplicates a section rather than dropping one.

## Fix

Match the sha anywhere in the heading rather than anchoring it to the end:

```bash
grep -qiE "^#{1,6}[[:space:]].*\b${skip_hash}\b" "$full"
```

`\b` keeps the property the end-anchor was reaching for — the #62 note explains the match is heading-only so a context bullet naming another commit's sha ("reverts deadbee") can't suppress that commit's own record — while allowing the position the writers actually use. Worth adding `-i` too, since a caller could pass an uppercase sha and the validator accepts `[0-9a-fA-F]`.

A test would want both positions (`## <sha> — msg` and `## msg <sha>`), a bullet-only mention that must *not* suppress, and a case-mismatched sha.

nhangen:
Fixed by #105, merged as `00781d9`. The PR title listed the issue numbers but the body had no `Closes` line, so GitHub did not close them automatically.

Issue #100, opened 2026-08-12
keeper append writes --section as plain text, not a heading


nhangen:
`keeper append --section "<title>"` writes the section title as a bare paragraph instead of a markdown heading, so appended sections don't appear in Obsidian's outline or document map.

Observed 2026-08-12 in `Projects/Development/nhangen/llm-tools/2026-08-12.md`. Sections written earlier in the day by the commit-capture path render as `## HH:MM — <title> <sha>` and appear in the outline. Sections appended later in the same session via direct `keeper append --section` calls landed as plain text at lines 533 and 557 — content intact, tables intact, but invisible to navigation.

Reproduce:

```
keeper append --vault <v> --target <note> --section "Some title" --body-file body.md
grep -n '^#' <note>   # the new title is absent
```

Expected: the section title is emitted as a heading at a consistent level (`##`), matching what the commit-capture path produces, so appended and captured sections are navigationally equivalent.

Impact is low but compounding: a long dated note accumulates sections that can't be jumped to, and the `--skip-if-hash` dedup regex anchors on a heading shape (`^#{1,6}[[:space:]]...`). If a section is written without a heading, a future append with the same hash may not detect it as already present — worth checking whether the plain-text path can defeat dedup.

nhangen:
Fixed by #105, merged as `00781d9`. The PR title listed the issue numbers but the body had no `Closes` line, so GitHub did not close them automatically.

PR #105, opened 2026-08-30
Clear the open issue list, easiest first (#94, #95, #100, #101, #103, #104)

nhangen:
All six open issues, worked easiest to hardest, plus the gaps three review rounds turned up. Every ticket is a silent failure — something that fails in a direction nobody hears — so each fix has coverage that fails without it.

## #101 — `--skip-if-hash` could never match the heading commit-capture writes

The gate anchored the sha to the **end** of a heading, but every commit capture writes it at the start (`## 7ac9556 — msg`), so the only guard against a double capture never fired. It now matches the sha anywhere in a heading, with hex-class boundaries (BSD grep/awk carry no `\b`) and a case-folded compare, since the validator accepts `[0-9a-fA-F]`. Heading-only is preserved — a context bullet naming another commit's sha must not suppress that commit's own record.

The match runs in `awk` rather than `grep -E '^#' | grep -q`: `-q`'s early exit SIGPIPEs the upstream grep, which `set -o pipefail` then reports as a failed match.

## #104 — `keeper insert` swallowed an INDEX-link failure via `|| true`

The note is written before the link step, so a failed link still printed the path and exited 0: the caller relayed success for a note nothing linked. `insert` now verifies the link after the apply and warns on stderr naming the note and the INDEX. Checking `vault_index_apply`'s rc alone would miss the common shape — a coverage defect is only reported on its stderr and still returns 0. Exit status stays 0, because the note itself did commit.

## #100 — `keeper append` wrote `--section` as plain text

A bare title landed as a paragraph: content intact, but absent from Obsidian's outline — and invisible to the heading-shaped `--skip-if-hash` gate above, so the same commit could capture twice. A bare value is now promoted to a level-2 heading; a value that already carries its own `#` prefix is kept verbatim so the caller picks the level; a whitespace-only value falls back to the default heading rather than emitting an empty one.

## #103 — the `dedup_jaccard_threshold` override reached no call site

The full save-conversation procedure interpolated a shell variable nothing ever set, and the librarian passed no threshold at all, so both ran at the library's 0.4 while the vault said otherwise. Resolved in the scanner instead of at each call site, since that is what the call sites kept failing to do: `dedup_same_day` reads the setting when given no threshold argument, via a new `obsidian_config_value` reader in `resolve-config.sh`. An explicit argument still wins — and is validated the same way, so a malformed one cannot reach `awk`, where a non-numeric threshold compares as 0 and matches the first note in the folder.

## #94 — epoch state files were read unvalidated

A zero-byte file — what ENOSPC leaves, since `>` truncates on open and can then fail on write — read as absent: an empty `last_attempt` made the banner take the never-attempted branch and stay quiet on a host that has attempted and never completed, and the mirror case printed "index has never completed a scan" on a host that just completed one.

One validated reader now serves both, separating the three states the banner has to distinguish rather than collapsing them onto emptiness: rc 0 with the value, rc 1 for no file, rc 2 for a file holding no timestamp. `staleness_banner` gets its third branch for that last state. Writes stage into the target's own directory and swap, so a failed write leaves the previous timestamp rather than a zero-byte file — and the `mv` is a rename, not a cross-filesystem copy. `ask-staleness.sh` no longer swallows its own failure, and no longer aborts before reaching it.

## #95 — a deferring host had no signal that the owner's scans are faulting

Not fixed by hoisting `keeper_record_attempt` above the ownership gate: that makes a deferring host claim an attempt it never made and warn forever about a vault the owner is scanning fine, which `tests/vaultkeeper-tick.sh` pins. The evidence is already in the vault and the banner ignored it. Where the host has no local state at all, it now reads what is synced — `Librarian.md`, which the owner stamps `scan_status: INCOMPLETE` into on every faulted tick, and the lease dir, which names the owner.

Four states, three of them a line: no claim by any host means the keeper has never run anywhere and stays quiet; a claim with no digest means no host has completed a scan; `INCOMPLETE` means the owner is faulting; and a digest nobody has refreshed past the same 2× grace the local branch allows is what a dead owner leaves, since it never gets to stamp `INCOMPLETE`. A healthy digest stays silent, and a host with local state is unaffected. The owner is named through `keeper_elect` rather than a second implementation of election.

## One bug found outside the tickets

`file_mtime` probed the two `stat` spellings by exit status: try BSD `-f %m`, fall back if it produced nothing. GNU `stat` reads `-f` as "report the filesystem" and **succeeds**, so on Linux every caller got a block of filesystem stats back as an mtime — `vault_index_plan` compared it with `-gt`, errored once per file, and silently hashed every note it walked, which is the entire cost the mtime gate exists to avoid. It now validates the answer instead of trusting the status.

## What review changed

**Round 1** — two of the six fixes did not reach a user as first written. The Codex package carries byte-identical copies of `scripts/keeper` and the libs it sources and was not re-vendored, so Codex users kept #100 and #101. And `ask-staleness.sh` aborted before it ever reached the banner: its config reads were one unmatched grep away from a failed pipeline under `pipefail`, and `/obsidian:setup`'s template never writes `keeper_interval_secs`, so on a default install the script exited 1 with no stdout — which every consumer reads as a healthy index. Also: save-conversation's steps 3 and 6 still gated on the literal `0.4`; `dedup_same_day`'s best-so-far seed made the documented `0.0` endpoint unreachable; `insert` ran a redundant recursive `dup_leaves` sweep.

**Round 2** (`f60521e`) — the Daily session link left a stray `.ktmp` on failure, and since the note commits first, the retry then died on "target already exists". `obsidian_config_value` scanned past the frontmatter and kept YAML quotes. The fast path hand-rolled config parsing and passed an unexpanded `<vault_path>`. And two tests were passing for the wrong reason: `tests/note-hash.sh` pinned the host's own `stat`, so on macOS it would have passed with the broken `file_mtime`; `tests/keeper-cli.sh` I4b was asserting on the overwrite guard's early abort rather than the dedup it claimed to cover.

**Round 3** (`7cd0afb`) — the round-2 hardening stopped three lines short. With `Daily/` unwritable and no daily note for that date, the `mkdir`, the note write and the `## Session Links` append were still bare, so `insert` exited 1 *after* committing the note, silently — and a caller reads non-zero as "not captured". The round-2 test pre-created the daily note, which is the one fixture shape that skips those lines. Also in this round:

- `ask-staleness.sh`'s `cfg_val` re-introduced the parsing defects `obsidian_config_value` exists to remove, one line below that function. A quoted `vault_path` kept its quotes, so `keeper_vault_health` found no lease dir and took its "stay quiet" branch — #95 undone for exactly the config shape a careful user writes. `session-save.sh` had the same read on the autosave path.
- That reader now strips a trailing ` # comment` as well as quotes, which is what made adopting it safe: `agents/vault-organizer.md` already stripped comments, so routing readers through a function that didn't would have regressed it. The strip requires whitespace before the `#` (so `/notes/#inbox` survives) and is skipped inside quoted scalars.
- `dedup_threshold` fell back to `0.4` in silence when its sibling `resolve-config.sh` was missing, while the unusable-value path beside it warns loudly. It now names the missing file, in the wording `allowlist-validate.sh` and `vault-scan.sh` use.
- `vaultkeeper-tick.sh` called `keeper_record_scan` bare while `keeper-state.sh` documented it as checked. Guarded, so the comment is true.
- save-conversation's `VAULT=` was unguarded: an empty vault scans `/<target-folder>`, `find`'s `2>/dev/null` makes that "no output", and the next line says no output means proceed — a config fault becomes "no duplicate found".
- Case 7's decoy sat *below* the real key, so awk's first-match exit won on file order and the frontmatter bound was never exercised — deleting the bound left it 7/7. A key present only in the prose body now pins it.

## Test notes

Suites are green here except eight that are red on `master` too, for reasons of this container rather than the code: assertions that need a non-root user (0444/0555 mean nothing to root), an `ulimit -n 64` fd-exhaustion probe, a missing `crontab` and a missing `codex` CLI, and the two `commit-capture` suites, which resolve no config of their own and depend on the developer's `~/.config/claude-obsidian/obsidian.local.md`. `note-hash` and `vault-index-plan` were in that red set and are now green, because their cause was the real bug above.

Four assertions skip under root with a printed reason; each was verified by running the suite as `nobody`, where it fails without its fix. Every round-3 fix was reproduced before the change and mutation-tested after.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01VsJcJR3RiUYjiVq8aUMS2S

### DIFF
--- a/scripts/ask-staleness.sh
+++ b/scripts/ask-staleness.sh
@@ -5,6 +5,9 @@ set -euo pipefail
 ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 . "${ROOT}/lib/note-hash.sh"
 . "${ROOT}/lib/keeper-state.sh"
+# Sourced so the banner can name the elected owner when it has to speak from
+# the vault's evidence rather than this host's (#95).
+. "${ROOT}/lib/keeper-lease.sh"
 . "${ROOT}/lib/resolve-config.sh"
 
 CONFIG="$(resolve_obsidian_config "${CLAUDE_PLUGIN_ROOT:-${ROOT%/scripts}}")" || exit 0
@@ -12,7 +15,10 @@ VAULT="$(grep '^vault_path:' "$CONFIG" | head -1 | sed 's/vault_path: *//')"
 [ -n "$VAULT" ] || exit 0
 INTERVAL="$(grep '^keeper_interval_secs:' "$CONFIG" | head -1 | sed 's/.*: *//')"
 INTERVAL="${INTERVAL:-900}"
+# `|| true`: the key is optional, and an unmatched grep is a failed pipeline
+# under this script's pipefail — which would abort the freshness check itself.
+PRIORITY="$(grep '^keeper_host_priority:' "$CONFIG" | head -1 | sed 's/.*: *//' | tr ' ' ',' || true)"
 # Do not let the check's own failure read as a clean bill of health: every
 # consumer keys off stdout alone ("if it prints a line, show it"), and this
 # script's set -e would otherwise exit 1 with nothing printed (#94).
-staleness_banner "$VAULT" "$INTERVAL" || printf '⚠ could not determine index freshness\n'
+staleness_banner "$VAULT" "$INTERVAL" "$PRIORITY" || printf '⚠ could not determine index freshness\n'
--- a/scripts/ask-staleness.sh
+++ b/scripts/ask-staleness.sh
@@ -11,13 +11,18 @@ ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 . "${ROOT}/lib/resolve-config.sh"
 
 CONFIG="$(resolve_obsidian_config "${CLAUDE_PLUGIN_ROOT:-${ROOT%/scripts}}")" || exit 0
-VAULT="$(grep '^vault_path:' "$CONFIG" | head -1 | sed 's/vault_path: *//')"
+# Every one of these is an unmatched-grep away from a failed pipeline under this
+# script's pipefail, and an abort here prints nothing — which every consumer
+# reads as a healthy index (#94). keeper_interval_secs is the live case: the
+# setup wizard's config template does not write it, so on a default install the
+# freshness check aborted one line before the guard meant to protect it, and the
+# `[ -n "$VAULT" ]` test below was dead code.
+cfg_val() { grep "^$1:" "$CONFIG" | head -1 | sed "s/^$1:[[:space:]]*//" || true; }
+VAULT="$(cfg_val vault_path)"
 [ -n "$VAULT" ] || exit 0
-INTERVAL="$(grep '^keeper_interval_secs:' "$CONFIG" | head -1 | sed 's/.*: *//')"
+INTERVAL="$(cfg_val keeper_interval_secs)"
 INTERVAL="${INTERVAL:-900}"
-# `|| true`: the key is optional, and an unmatched grep is a failed pipeline
-# under this script's pipefail — which would abort the freshness check itself.
-PRIORITY="$(grep '^keeper_host_priority:' "$CONFIG" | head -1 | sed 's/.*: *//' | tr ' ' ',' || true)"
+PRIORITY="$(cfg_val keeper_host_priority | tr ' ' ',')"
 # Do not let the check's own failure read as a clean bill of health: every
 # consumer keys off stdout alone ("if it prints a line, show it"), and this
 # script's set -e would otherwise exit 1 with nothing printed (#94).
--- a/scripts/ask-staleness.sh
+++ b/scripts/ask-staleness.sh
@@ -11,18 +11,18 @@ ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 . "${ROOT}/lib/resolve-config.sh"
 
 CONFIG="$(resolve_obsidian_config "${CLAUDE_PLUGIN_ROOT:-${ROOT%/scripts}}")" || exit 0
-# Every one of these is an unmatched-grep away from a failed pipeline under this
-# script's pipefail, and an abort here prints nothing — which every consumer
-# reads as a healthy index (#94). keeper_interval_secs is the live case: the
-# setup wizard's config template does not write it, so on a default install the
-# freshness check aborted one line before the guard meant to protect it, and the
-# `[ -n "$VAULT" ]` test below was dead code.
-cfg_val() { grep "^$1:" "$CONFIG" | head -1 | sed "s/^$1:[[:space:]]*//" || true; }
-VAULT="$(cfg_val vault_path)"
+# obsidian_config_value, not a local grep/sed: an abort here prints nothing,
+# which every consumer reads as a healthy index (#94) — and a hand-rolled read
+# keeps the quotes, so a config with `vault_path: "/vault"` sent a $VAULT that
+# matches no directory into keeper_vault_health, which then took its
+# "no host has ever claimed this vault, stay quiet" branch and hid a faulting
+# owner. Same shape one layer on for a quoted keeper_interval_secs.
+# `|| true`: an absent key returns 1, and each has a default below.
+VAULT="$(obsidian_config_value vault_path || true)"
 [ -n "$VAULT" ] || exit 0
-INTERVAL="$(cfg_val keeper_interval_secs)"
+INTERVAL="$(obsidian_config_value keeper_interval_secs || true)"
 INTERVAL="${INTERVAL:-900}"
-PRIORITY="$(cfg_val keeper_host_priority | tr ' ' ',')"
+PRIORITY="$(obsidian_config_value keeper_host_priority | tr ' ' ',' || true)"
 # Do not let the check's own failure read as a clean bill of health: every
 # consumer keys off stdout alone ("if it prints a line, show it"), and this
 # script's set -e would otherwise exit 1 with nothing printed (#94).
--- a/scripts/ask-staleness.sh
+++ b/scripts/ask-staleness.sh
@@ -12,4 +12,7 @@ VAULT="$(grep '^vault_path:' "$CONFIG" | head -1 | sed 's/vault_path: *//')"
 [ -n "$VAULT" ] || exit 0
 INTERVAL="$(grep '^keeper_interval_secs:' "$CONFIG" | head -1 | sed 's/.*: *//')"
 INTERVAL="${INTERVAL:-900}"
-staleness_banner "$VAULT" "$INTERVAL"
+# Do not let the check's own failure read as a clean bill of health: every
+# consumer keys off stdout alone ("if it prints a line, show it"), and this
+# script's set -e would otherwise exit 1 with nothing printed (#94).
+staleness_banner "$VAULT" "$INTERVAL" || printf '⚠ could not determine index freshness\n'
--- a/scripts/keeper
+++ b/scripts/keeper
@@ -207,7 +207,11 @@ cmd_insert() {
   local link_rc=0 stem
   vault_index_apply "$folder" "$idx" >/dev/null || link_rc=$?
   stem="$(basename "$target" .md)"
-  if ! vault_index_has_link "$idx" "$stem" "$(vault_index_dup_leaves "$folder")"; then
+  # No dup_leaves argument: $stem is a bare basename, so vault_index_has_link
+  # returns before it consults that set (the leaf IS the rel). Passing it meant
+  # a second full recursive sweep of the folder per insert, duplicating the one
+  # vault_index_apply just did.
+  if ! vault_index_has_link "$idx" "$stem"; then
     printf 'keeper: warning — %s is written but is not linked from %s%s\n' \
       "$target" "${idx#"$vault_abs"/}" \
       "$([ "$link_rc" -eq 0 ] || printf ' (INDEX step failed, rc=%s)' "$link_rc")" >&2
--- a/scripts/keeper
+++ b/scripts/keeper
@@ -220,9 +220,27 @@ cmd_insert() {
   # Optional: link the note into Daily/<date> under ## Session Links.
   if [ -n "$sldate" ]; then
     local daily="$vault_abs/Daily/$sldate.md"
-    mkdir -p "$vault_abs/Daily"
-    [ -f "$daily" ] || printf -- '---\ndate: %s\ntags: [daily]\n---\n\n# %s\n' "$sldate" "$sldate" > "$daily"
-    grep -q '^## Session Links' "$daily" || printf '\n## Session Links\n' >> "$daily"
+    # Every write below is guarded, not just the swap at the end. Bare under
+    # `set -e` these abort the whole insert AFTER the note has committed, with
+    # nothing on stderr — and a caller reads a non-zero exit as "not captured",
+    # so a capture that did happen is reported as lost and the retry then dies
+    # on "target already exists". The note is written; a daily link that cannot
+    # be made is a warning, not a failure.
+    local sl_warn="keeper: warning — $target is written but Daily/$sldate.md"
+    if ! mkdir -p "$vault_abs/Daily" 2>/dev/null; then
+      printf '%s could not be created; no session link\n' "$sl_warn" >&2
+      printf '%s\n' "$full"; return 0
+    fi
+    if [ ! -f "$daily" ] \
+       && ! printf -- '---\ndate: %s\ntags: [daily]\n---\n\n# %s\n' "$sldate" "$sldate" > "$daily" 2>/dev/null; then
+      printf '%s could not be created; no session link\n' "$sl_warn" >&2
+      printf '%s\n' "$full"; return 0
+    fi
+    if ! grep -q '^## Session Links' "$daily" \
+       && ! printf '\n## Session Links\n' >> "$daily" 2>/dev/null; then
+      printf '%s is unwritable; no session link\n' "$sl_warn" >&2
+      printf '%s\n' "$full"; return 0
+    fi
     # Link the note by its vault-relative path so it resolves from Daily/, with
     # the human title as an alias. Dedup on the path: two different notes can
     # share a title, and keying on title text dropped the second one's link.
--- a/scripts/keeper
+++ b/scripts/keeper
@@ -191,8 +191,14 @@ cmd_insert() {
   # differed from the stem produced two entries, the title one resolving to
   # nothing. Keep stderr: apply's coverage report is the only signal that the
   # slice is under-indexed.
-  local idx="$folder/INDEX.md"
-  [ -f "$idx" ] || printf '# %s Index\n' "$(basename "$folder")" > "$idx"
+  local idx="$folder/INDEX.md" iprep=1
+  # Guarded for the same reason as the Daily/ writes below: the note committed
+  # above, so aborting here under `set -e` reports "not captured" for a capture
+  # that happened. An INDEX.md that cannot be created (ENOSPC, or a directory
+  # sitting on the name) is a link failure, and a link failure is a warning.
+  if [ ! -f "$idx" ]; then
+    printf '# %s Index\n' "$(basename "$folder")" > "$idx" 2>/dev/null || iprep=0
+  fi
   . "$KEEPER_DIR/lib/note-hash.sh"
   . "$KEEPER_DIR/lib/vault-index.sh"
   # `|| true` alone made a failed link indistinguishable from a written one
@@ -205,13 +211,13 @@ cmd_insert() {
   # apply's stderr and still returns 0, so rc by itself would miss the common
   # shape (an unwritable INDEX.md).
   local link_rc=0 stem
-  vault_index_apply "$folder" "$idx" >/dev/null || link_rc=$?
+  [ "$iprep" = 1 ] && { vault_index_apply "$folder" "$idx" >/dev/null || link_rc=$?; }
   stem="$(basename "$target" .md)"
   # No dup_leaves argument: $stem is a bare basename, so vault_index_has_link
   # returns before it consults that set (the leaf IS the rel). Passing it meant
   # a second full recursive sweep of the folder per insert, duplicating the one
   # vault_index_apply just did.
-  if ! vault_index_has_link "$idx" "$stem"; then
+  if [ "$iprep" = 0 ] || ! vault_index_has_link "$idx" "$stem"; then
     printf 'keeper: warning — %s is written but is not linked from %s%s\n' \
       "$target" "${idx#"$vault_abs"/}" \
       "$([ "$link_rc" -eq 0 ] || printf ' (INDEX step failed, rc=%s)' "$link_rc")" >&2
--- a/scripts/keeper
+++ b/scripts/keeper
@@ -42,6 +42,23 @@ keeper_guard_target() {
   esac
 }
 
+# Does any section heading in $1 name the sha $2? Heading-only (a context bullet
+# legitimately names other commits' shas) and position-independent: writers put
+# the sha at the head of the line ("## <sha> — msg") as often as at the end, and
+# an end-anchored match never fired for them (#101).
+#
+# awk, not a grep pipeline: `grep -E '^#' | grep -q` lets -q's early exit SIGPIPE
+# the first grep, which `set -o pipefail` then reports as a failed match.
+# Boundaries are hex-class rather than \b — BSD grep/awk do not carry \b — and
+# the compare is case-folded because the validator accepts [0-9a-fA-F].
+keeper_heading_has_sha() {
+  awk -v sha="$2" '
+    BEGIN { sha = tolower(sha); pat = "(^|[^0-9a-f])" sha "([^0-9a-f]|$)" }
+    $1 ~ /^##?#?#?#?#?$/ && NF > 1 && tolower($0) ~ pat { found = 1; exit }
+    END { exit(found ? 0 : 1) }
+  ' "$1"
+}
+
 # append a timestamped section to a dated note, creating it on absence.
 cmd_append() {
   local vault="" target="" date="" section="" body_file="" init_file="" skip_hash="" skip_given=0
@@ -84,7 +101,7 @@ cmd_append() {
       '') keeper_die "append: --skip-if-hash requires a value" ;;
       *[!0-9a-fA-F]*) keeper_die "append: --skip-if-hash must be a hex sha: $skip_hash" ;;
       *)
-        if [ -f "$full" ] && grep -qE "^#{1,6}[[:space:]].*[[:space:]]${skip_hash}[[:space:]]*\$" "$full"; then
+        if [ -f "$full" ] && keeper_heading_has_sha "$full" "$skip_hash"; then
           printf 'keeper: append skipped — %s already has a section for %s\n' "$target" "$skip_hash" >&2
           printf '%s\n' "$full"
           return 0
@@ -165,7 +182,23 @@ cmd_insert() {
   [ -f "$idx" ] || printf '# %s Index\n' "$(basename "$folder")" > "$idx"
   . "$KEEPER_DIR/lib/note-hash.sh"
   . "$KEEPER_DIR/lib/vault-index.sh"
-  vault_index_apply "$folder" "$idx" >/dev/null || true
+  # `|| true` alone made a failed link indistinguishable from a written one
+  # (#104): the note is committed before this runs, so insert still printed the
+  # path and exited 0 while nothing in the vault pointed at the note. Keep the
+  # exit status (the note IS written — failing here would tell the caller the
+  # opposite), but say on stderr what is missing.
+  #
+  # Check the link, not just apply's rc: a coverage defect is reported on
+  # apply's stderr and still returns 0, so rc by itself would miss the common
+  # shape (an unwritable INDEX.md).
+  local link_rc=0 stem
+  vault_index_apply "$folder" "$idx" >/dev/null || link_rc=$?
+  stem="$(basename "$target" .md)"
+  if ! vault_index_has_link "$idx" "$stem" "$(vault_index_dup_leaves "$folder")"; then
+    printf 'keeper: warning — %s is written but is not linked from %s%s\n' \
+      "$target" "${idx#"$vault_abs"/}" \
+      "$([ "$link_rc" -eq 0 ] || printf ' (INDEX step failed, rc=%s)' "$link_rc")" >&2
+  fi
 
   # Optional: link the note into Daily/<date> under ## Session Links.
   if [ -n "$sldate" ]; then
--- a/scripts/keeper
+++ b/scripts/keeper
@@ -121,7 +121,20 @@ cmd_append() {
     fi
   fi
 
+  # --section names a section, so it is written as a heading (#100). Callers
+  # that pass a bare title ("Some title") had it land as a paragraph: content
+  # intact, but absent from Obsidian's outline, and invisible to the
+  # heading-shaped --skip-if-hash gate above, which would then let the same
+  # commit be captured twice. A value that already carries its own `#` prefix
+  # is kept verbatim, so a caller can choose the level.
+  case "$section" in *[![:space:]]*) : ;; *) section="" ;; esac   # blanks == none
   [ -n "$section" ] || section="## $(date +%H:%M) — entry"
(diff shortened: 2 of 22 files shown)


scripts/keeper
```diff
@@ -165,7 +195,27 @@ cmd_insert() {
...
+  # a second full recursive sweep of the folder per insert, duplicating the one
+  # vault_index_apply just did.
+  if ! vault_index_has_link "$idx" "$stem"; then
+    printf 'keeper: warning — %s is written but is not linked from %s%s\n' \
+      "$target" "${idx#"$vault_abs"/}" \
+      "$([ "$link_rc" -eq 0 ] || printf ' (INDEX step failed, rc=%s)' "$link_rc")" >&2
+  fi
 
   # Optional: link the note into Daily/<date> under ## Session Links.
   if [ -n "$sldate" ]; then
```
nhangen, line 227: A bit further down around line 238, the Daily note linking runs `awk ... > "$daily.ktmp" && mv "$daily.ktmp" "$daily"`. If that `mv` fails (such as read-only permissions or disk hiccup), under `set -e` it aborts the process without cleaning up `$daily.ktmp`. That leaves a stray `.ktmp` in the user's `Daily/` folder, and because the note itself was already committed at line 183, a subsequent retry fails with `insert: target already exists`. Staging with `keeper_swap_or_clean` and warning on stderr (preserving exit 0 like the index link above) would match the rest of `cmd_insert`.

scripts/lib/dedup-scan.sh
```diff
@@ -32,14 +32,52 @@ jaccard() {
...
+  fi
+  printf 'dedup_same_day: unusable dedup_jaccard_threshold %s — using %s\n' \
+    "$v" "$DEDUP_JACCARD_DEFAULT" >&2
+  printf '%s\n' "$DEDUP_JACCARD_DEFAULT"
+}
+
 dedup_same_day() {
-  local folder="$1" date="$2" slug="$3" threshold="${4:-0.4}"
+  local folder="$1" date="$2" slug="$3" threshold="${4:-}"
+  [ -n "$threshold" ] || threshold="$(dedup_threshold)"
```
nhangen, line 88: When `$4` is passed explicitly, `dedup_threshold()` is bypassed. If a caller passes a non-numeric string (or quoted float), it reaches `awk -v t="$threshold"`, where non-numeric values coerce to 0 and match the first note in the folder. Validating `$threshold` inside `dedup_same_day` even when supplied as an argument avoids letting malformed strings reach awk.

scripts/lib/resolve-config.sh
```diff
@@ -40,6 +40,22 @@ resolve_obsidian_config() {
...
+# the trimmed value; returns 1 when no config resolves or the key is absent, so
+# a caller can tell "unset" (use my default) from "set". Config keys documented
+# in a SKILL and written by setup were otherwise read by whichever call site
+# remembered to (#103) — dedup_jaccard_threshold reached none of its three.
+obsidian_config_value() {
+  local key="$1" cfg v
+  cfg="$(resolve_obsidian_config "${2:-${CLAUDE_PLUGIN_ROOT:-}}")" || return 1
+  v="$(awk -v k="$key" '
+    index($0, k ":") == 1 {
+      sub(/^[^:]*:[[:space:]]*/, ""); sub(/[[:space:]]+$/, ""); print; exit
```
nhangen: Surrounding quotes aren't stripped here, so quoted scalars like `"0.2"` retain quotes when returned. Stripping leading/trailing quotes here would let quoted numbers in YAML parse cleanly.

tests/note-hash.sh
```diff
@@ -29,4 +29,18 @@ note_hash_valid "" && fail "should reject empty"
 M="$(file_mtime "$TMP/a.md")"; [[ "$M" =~ ^[0-9]+$ ]] || fail "file_mtime not integer: $M"
 N="$(now_epoch)"; [[ "$N" =~ ^[0-9]+$ ]] || fail "now_epoch not integer: $N"
 
+# The two spellings cannot be told apart by exit status: GNU stat reads `-f` as
+# a filesystem query and SUCCEEDS on it, returning a block of filesystem stats.
+# Trusting that answer gave every caller a non-numeric mtime — vault_index_plan
+# then errored per file and silently hashed every note it walked.
+MT="$(file_mtime "$TMP/a.md")"
```
nhangen, line 36: On macOS/BSD systems, `stat -f %m` natively succeeds with an integer epoch. Reverting `file_mtime` to the master branch implementation still passes this test because both lines 29 and 36 exercise host BSD stat. Stubbing `stat` in a subshell to return multiline filesystem info on `-f %m` would verify that `file_mtime` actually falls back to `-c %Y`.

tests/keeper-cli.sh
```diff
@@ -188,6 +248,35 @@ grep -qxF -- '- [[Notes/plain]]' "$V/Notes/INDEX.md" || fail "default title (no
 bash "$KEEPER" insert --vault "$V" --target "../evil.md" --body-file "$TMP/note.md" 2>/dev/null && fail "insert path traversal must fail"
```
nhangen, line 254: In test I4b (earlier around line 227), re-running `insert` on the identical note hits the overwrite guard at line 181 (`insert: target already exists`) and aborts before ever reaching the Session Links deduplication logic at line 221. The test passes today purely because of the early abort. Pre-populating `$DLY` with a pre-existing link before the insert would let it test the dedup check itself.

nhangen (commented):
Self-authored PR override: downgraded review event to COMMENT per GitHub constraint.

Mini panel review covering silent failure modes, test invariants, and plugin conventions across all six ticket fixes. Notes on daily note tempfile cleanup, explicit threshold validation, GNU stat test simulation, and test I4b are anchored inline.

Additional observations:
- `obsidian_config_value` has no direct unit tests in `tests/resolve-config.sh`, and currently scans across the full file rather than bounding between `---` frontmatter markers.
- `skills/save-conversation/SKILL.md` fast path still manually parses `THRESHOLD` from config and carries an unexpanded `<vault_path>` placeholder.
- In `tests/dedup-scan.sh:57`, environment scrubbing for the unconfigured test case omits `OBSIDIAN_LOCAL_MD` and `OBSIDIAN_DEDUP_JACCARD_THRESHOLD`.

scripts/keeper
nhangen: Fixed in f60521e. Staged through `keeper_swap_or_clean` with a `mktemp` in `Daily/`, warning on stderr and keeping exit 0, matching the INDEX link above.

One correction while testing it: the reachable failure is a read-only `Daily/` **directory**, not a read-only note. `rename(2)` only needs write permission on the directory, so a 0444 daily note is replaced without complaint — my first attempt at the test injected the failure that way and never triggered anything. New test I11 makes the directory unwritable, and asserts exit 0, the note still written, a warning naming the daily note, and zero `*ktmp*` files left in `Daily/`. It fails against the old `> "$daily.ktmp" && mv` (silent, and the stray file is exactly what it leaves).

---
_Generated by [Claude Code](https://claude.ai/code)_

scripts/lib/dedup-scan.sh
nhangen: Fixed in f60521e. The validation moved into a `dedup_threshold_valid` predicate that both paths use, so an argument is checked exactly like a resolved value: a bad one warns on stderr and falls back to the resolved default rather than reaching awk.

Covered in `tests/dedup-scan.sh` for both shapes you named — `not-a-number` and a quoted `"0.2"`. Worth noting the quoted case does not become 0.2; it is rejected and the default applies. Stripping quotes belongs in the config reader (where YAML put them), not in an argument a caller controls.

---
_Generated by [Claude Code](https://claude.ai/code)_

scripts/lib/resolve-config.sh
nhangen: Fixed in f60521e — quotes stripped, and the reader is now bounded to the frontmatter block as your review body suggested, so prose below it that opens a line with `vault_path:` can't answer first.

The bound is conditional: if line 1 isn't `---`, it scans the whole file as before. A strict bound would make a config with a leading blank line read as empty for every key, which is worse than the over-broad scan it replaces.

`tests/resolve-config.sh` case 7 now covers this directly (it had no coverage at all): quoted scalar, whitespace trimming, a decoy in the prose, an absent key returning 1, and a config with no frontmatter markers.

---
_Generated by [Claude Code](https://claude.ai/code)_

tests/note-hash.sh
nhangen: You're right, and this was the most useful catch of the review: the test was pinning the host's `stat`, so on the platform this plugin actually targets it would have passed with the broken implementation.

Fixed in f60521e with the stub you describe — `-f` exits 0 with multiline filesystem stats, `-c %Y` returns an epoch — asserting `file_mtime` returns the epoch. Verified it separates the two implementations rather than just passing:

```
--- master file_mtime under the stub:
  File: "x"
    ID: 0        Namelen: 255
--- branch file_mtime under the same stub:
1712345678
```

Added a second stub where neither spelling answers, to pin that it fails loudly instead of echoing something the caller does arithmetic on.

---
_Generated by [Claude Code](https://claude.ai/code)_

tests/keeper-cli.sh
nhangen: Confirmed and fixed in f60521e. I4b was asserting on the overwrite guard's early abort, so the `grep -qF -- "[[$ltarget]]"` dedup it claimed to cover was never reached — the `&& fail "insert must refuse to overwrite; I4b cannot test dedup this way"` line was even admitting it.

It now seeds the link into `$DLY` first (the state a Syncthing round-trip or a hand-edit leaves), then inserts the note that link points at, so the insert runs to completion and the dedup branch decides the outcome. Asserts the link appears exactly once, with a fixture precondition check so a broken seed can't make it pass vacuously.

---
_Generated by [Claude Code](https://claude.ai/code)_

nhangen:
All five inline threads and the three observations from the review body are addressed in f60521e. The inline ones are answered and resolved on their threads; the three from the body:

**`obsidian_config_value` had no direct tests and scanned the whole file.** Both fixed. It is now bounded to the frontmatter when the file opens with `---`, and falls back to the full-file scan when it doesn't — a strict bound would make a config with a leading blank line read as empty for *every* key, which is a worse failure than the one it fixes. `tests/resolve-config.sh` case 7 covers the bound, quote stripping, trimming, an absent key, and the no-frontmatter fallback.

**save-conversation's fast path re-parsed `THRESHOLD` and carried an unexpanded `<vault_path>`.** Both were live bugs, not just untidiness: the placeholder means the documented command resolves a path that doesn't exist, and the re-parse is the same "call site decides the threshold" pattern #103 exists to remove — it would have gone on disagreeing with the scanner. It now sources `resolve-config.sh` for `vault_path` and passes no threshold, letting `dedup_same_day` resolve it. The pin in `tests/sole-brain-save-conversation.sh` asserted the old hand-rolled shape, so it now pins the contract instead, including that no `<vault_path>` placeholder survives in a runnable command.

**`tests/dedup-scan.sh` scrub was incomplete.** `OBSIDIAN_LOCAL_MD` and `OBSIDIAN_DEDUP_JACCARD_THRESHOLD` are scrubbed too — either one set in a developer's shell answers ahead of everything that case pins.

Suite state is unchanged from before this round: the same eight suites are red here as on `master`, all for container reasons (root ignoring 0444/0555, an `ulimit -n` probe, missing `crontab` and `codex` CLIs, and the two `commit-capture` suites depending on a developer's `~/.config/claude-obsidian/obsidian.local.md`). The two new non-root assertions were verified by running the suite as `nobody`, where each fails without its fix.

---
_Generated by [Claude Code](https://claude.ai/code)_

tests/resolve-config.sh
```diff
@@ -75,4 +75,28 @@ GOT=$(env XDG_CONFIG_HOME="$XDG" bash "$RESOLVER" --stable)
 [ "$GOT" = "$STABLE" ] || fail "case6 (--stable): got '$GOT' want '$STABLE'"
 PASS_COUNT=$((PASS_COUNT + 1))
 
-printf '%d/6 cases passed\n' "$PASS_COUNT"
+# ----- Case 7: obsidian_config_value reads frontmatter scalars -----
```
nhangen, line 78: Case 7's decoy `vault_path: /decoy` sits *below* the real key, so awk's first-match `exit` wins on file order alone and the frontmatter bound is never exercised. Deleting the two bound lines in `resolve-config.sh` leaves this at 7/7 passing. A key present only in the prose body, asserted `rc=1`, would pin it.

scripts/lib/resolve-config.sh
```diff
@@ -40,6 +40,30 @@ resolve_obsidian_config() {
   return 1
 }
 
+# Read a scalar `key: value` out of the resolved config's frontmatter. Echoes
+# the trimmed value; returns 1 when no config resolves or the key is absent, so
+# a caller can tell "unset" (use my default) from "set". Config keys documented
+# in a SKILL and written by setup were otherwise read by whichever call site
+# remembered to (#103) — dedup_jaccard_threshold reached none of its three.
+obsidian_config_value() {
```
nhangen, line 48: `obsidian_config_value` doesn't strip a trailing `# comment` — `vault_path: /v  # my vault` comes back as `/v  # my vault`. `agents/vault-organizer.md:15` already hand-rolls both the comment strip and the quote strip, so routing every reader through this function (which is what the rest of the PR is heading toward) would regress that one and still wouldn't fix the comment case in `session-save.sh`. Worth moving comment-stripping in here first.

Separately, `[ -n "$v" ] || return 1` treats a key present with an empty value as absent. Harmless for today's two keys.

scripts/ask-staleness.sh
```diff
@@ -5,11 +5,25 @@ set -euo pipefail
...
 
 CONFIG="$(resolve_obsidian_config "${CLAUDE_PLUGIN_ROOT:-${ROOT%/scripts}}")" || exit 0
-VAULT="$(grep '^vault_path:' "$CONFIG" | head -1 | sed 's/vault_path: *//')"
+# Every one of these is an unmatched-grep away from a failed pipeline under this
+# script's pipefail, and an abort here prints nothing — which every consumer
+# reads as a healthy index (#94). keeper_interval_secs is the live case: the
+# setup wizard's config template does not write it, so on a default install the
+# freshness check aborted one line before the guard meant to protect it, and the
+# `[ -n "$VAULT" ]` test below was dead code.
+cfg_val() { grep "^$1:" "$CONFIG" | head -1 | sed "s/^$1:[[:space:]]*//" || true; }
```
nhangen: `resolve-config.sh` is already sourced at `:11`, so `obsidian_config_value` is in scope — and then `cfg_val` reintroduces both defects it was written to fix. Measured on a fixture with a claimed lease and an `INCOMPLETE` digest: an unquoted `vault_path` prints the faulted-scan warning, a quoted one prints nothing and exits 0. The quotes survive into `$VAULT`, so `keeper_vault_health` takes the "keeper never ran anywhere, stay quiet" branch and a faulting owner is indistinguishable from a healthy one. `keeper_interval_secs: "900"` has the same shape one layer on.

scripts/lib/dedup-scan.sh
```diff
@@ -32,14 +32,62 @@ jaccard() {
...
+# A float in 0.0-1.0. Anything else compares as 0 inside awk and matches the
+# first note in the folder, so every path into dedup_same_day is checked — the
+# argument one included, since a caller can pass a quoted or malformed value.
+dedup_threshold_valid() {
+  printf '%s' "$1" | grep -qE '^(0(\.[0-9]+)?|1(\.0+)?|\.[0-9]+)$'
+}
+
+dedup_threshold() {
+  local v="${OBSIDIAN_DEDUP_JACCARD_THRESHOLD:-}"
+  if [ -z "$v" ] && [ -n "$_ds_lib_dir" ] && [ -f "$_ds_lib_dir/resolve-config.sh" ]; then
```
nhangen: No `else` here, so a missing sibling `resolve-config.sh` yields the `0.4` default with nothing on stderr — a vault that set `0.0` runs at `0.4`, which is the twin-note failure #103 exists to prevent. `allowlist-validate.sh:43` and `vault-scan.sh:25` both name the missing file and say "the install is broken, not the config"; this is the one place that goes quiet. The value path just below already warns loudly, so the two halves disagree.

skills/save-conversation/SKILL.md
```diff
@@ -40,16 +40,18 @@ below; when in doubt, use the full procedure.
...
 ```bash
 CONFIG="$(bash "${CLAUDE_PLUGIN_ROOT}/scripts/lib/resolve-config.sh")"
-# read vault_path and dedup_jaccard_threshold from $CONFIG:
-VAULT="$(grep '^vault_path:' "$CONFIG" | head -1 | sed 's/^vault_path:[[:space:]]*//')"
-THRESHOLD="$(awk '/^dedup_jaccard_threshold:/ {print $2}' "$CONFIG")"
-THRESHOLD="${THRESHOLD:-0.4}"
+# One reader for config scalars: it bounds itself to the frontmatter and strips
+# YAML quotes, which hand-rolled greps here did not.
+. "${CLAUDE_PLUGIN_ROOT}/scripts/lib/resolve-config.sh"
+VAULT="$(obsidian_config_value vault_path)"
```
nhangen: Unguarded, where `allowlist_validate … || exit 0` three lines down has the idiom. With `VAULT=""` the next call scans `/<target-folder>`, `dedup_same_day`'s `find … 2>/dev/null` swallows the error and echoes nothing, and the snippet's own next line reads `# no output above → proceed` — so the agent files a duplicate note and nothing reports it.

nhangen (commented):
Mini panel (three specialists plus an auditor) on `f60521e`. Suite green, shellcheck clean, vendored `packages/codex` copies byte-identical to `scripts/`. Every claim was mutation-tested in a scratch copy.

Four of the five earlier review items are fixed and covered — reverting each breaks a named test. The quote-stripping item is half done: the fix landed in the new `obsidian_config_value`, but the two production readers still hand-roll the parse.

Three findings with no line in this diff:

- `scripts/keeper:223-225` — the daily-note hardening stops three lines short of the failure the earlier reply named. With `Daily/` unwritable and no daily note yet, `insert` exits 1 *after* the note is committed, prints no warning, and the retry dies on "target already exists". Callers read a non-zero exit as "not captured", so a capture that happened is reported as lost. `tests/keeper-cli.sh:296` pre-creates the daily note — the one fixture shape that skips lines 224 and 225. The vendored copy needs the same change.
- `scripts/session-save.sh:23-26` — the autosave Stop hook exits 0 on a config it can't parse. Pre-existing, but this PR builds the right reader, so it's the natural place to adopt it — after the comment-stripping gap noted inline is closed.
- `scripts/vaultkeeper-tick.sh:207` — `keeper_record_scan` is bare where its sibling at `:52` is guarded with a rationale paragraph. A failed write aborts the tick after every shared write has landed, and `keeper-state.sh:70`'s comment claims a check only `record_attempt` has.

Coverage gaps, none blocking: nothing pins the five vendored Codex files against `scripts/`; the `(INDEX step failed, rc=%s)` branch is unreachable by the suite; an out-of-range threshold like `10` falls back to the looser `0.4`, untested.

tests/resolve-config.sh
nhangen: Confirmed by reproducing your mutation before fixing it: deleting the two bound lines left case 7 at **7/7 passing**. The decoy was doing nothing.

Fixed in 7cd0afb with `prose_only_key`, which appears only below the closing `---` and is asserted `rc=1`. Same mutation now fails:

```
FAIL: case7: a key present only below the frontmatter resolved: '/also-decoy'
```

Added the comment cases alongside it — a trailing ` # comment`, a `#` that is part of the value, and a quoted scalar with a comment after it.

---
_Generated by [Claude Code](https://claude.ai/code)_

scripts/lib/resolve-config.sh
nhangen: Right, and the ordering argument is what made this worth doing first — adopting the reader elsewhere before it stripped comments would have *regressed* `agents/vault-organizer.md:15`, which strips both. Reproduced: `vault_path: /v  # my vault` came back as `/v  # my vault`.

Fixed in 7cd0afb. Two refinements over the hand-rolled version:

- The strip requires whitespace before the `#`, so a vault at `/notes/#inbox` keeps its name. `vault-organizer`'s `s/[[:space:]]*#.*//` would eat that.
- Inside a quoted scalar the `#` is literal, so quoted values are taken up to their closing quote and not comment-stripped: `"0.3"  # tuned down` → `0.3`.

That made it safe to adopt in `ask-staleness.sh` and `session-save.sh`, both in the same commit.

On the empty-value point — agreed it's harmless for today's keys, and I've left it. `[ -n "$v" ] || return 1` is what lets a caller use `|| true` with a default; distinguishing "present but empty" from "absent" would need a second channel, and no caller wants that distinction yet.

---
_Generated by [Claude Code](https://claude.ai/code)_

scripts/ask-staleness.sh
nhangen: Fair hit — I wrote `cfg_val` to fix the `pipefail` abort and reintroduced the parsing defects one line below the function that exists to prevent them. And the consequence you traced is the worse one: quotes survive into `$VAULT`, `keeper_vault_health` finds no lease directory, takes its "no host has ever claimed this vault, stay quiet" branch, and #95 is undone for exactly the config shape a careful user writes.

Fixed in 7cd0afb — all three reads (`vault_path`, `keeper_interval_secs`, `keeper_host_priority`) go through `obsidian_config_value`, with `|| true` since each has a default and an absent key returns 1. The `pipefail` protection the old function was for is inherent to the shared reader, which is a single `awk` rather than a `grep | head | sed` pipeline.

`scripts/session-save.sh` had the same hand-rolled read on the autosave path, where the guard below it turns an unparsed config into a silent exit 0 — the session simply never gets saved. Fixed in the same commit, which is the "after the comment-stripping gap is closed" you called for in the body.

---
_Generated by [Claude Code](https://claude.ai/code)_

scripts/lib/dedup-scan.sh
nhangen: Fixed in 7cd0afb, with the wording from `allowlist-validate.sh` / `vault-scan.sh` — the install is broken, not the config — naming the missing path and the default it fell back to.

The "two halves disagree" framing is the right diagnosis: the value path warns because a bad value is the user's to fix, and this path stayed quiet because a missing file felt like it wasn't. But the consequence is identical from the vault's side — a vault that set `0.0` runs at `0.4` and writes the twin note.

Covered by a test that copies `dedup-scan.sh` alone into an orphan directory (no sibling), with the resolver's env scrubbed, and asserts the message names `resolve-config.sh`. Reverting to the old `&&`-chained condition fails it:

```
FAIL: a missing resolve-config.sh sibling fell back to the default in silence: <empty>
```

Also added the out-of-range case from your coverage note — `10` warns and falls back rather than silently disabling the check.

---
_Generated by [Claude Code](https://claude.ai/code)_

skills/save-conversation/SKILL.md
nhangen: Fixed in 7cd0afb — `|| exit 0` with an inline note on why, matching the `allowlist_validate` idiom three lines down.

The chain you traced is the reason this one matters more than it looks: an empty `$VAULT` doesn't error, it scans `/<target-folder>`; `find`'s `2>/dev/null` turns that into no output; and the snippet's very next line tells the agent that no output means proceed. A config fault silently becomes "no duplicate found", which is the one answer that causes the write it was supposed to prevent.

---
_Generated by [Claude Code](https://claude.ai/code)_

nhangen:
All five inline threads plus the three no-line findings are fixed in 7cd0afb, each reproduced before the fix and mutation-tested after. Answering the three from the body:

**`scripts/keeper:223-225` — the hardening stopped three lines short.** This was the important one, and your diagnosis of my own test is exact: `tests/keeper-cli.sh:296` pre-created the daily note, which is the single fixture shape that skips the `mkdir`, the note write and the `## Session Links` append. With `Daily/` unwritable and no note for that date, `insert` exited 1 *after* committing the note, silently — and a caller reads non-zero as "not captured", so a capture that happened is reported lost and the retry dies on "target already exists". All three writes are guarded now; I11b covers the missing-note shape and fails against the previous commit with exactly that symptom:

```
FAIL: insert exited 1 after committing the note; a caller reads that as 'not captured'
```

The vendored copy is re-synced.

**`scripts/session-save.sh:23-26`.** Adopted, in the same commit as the comment-stripping — which was the precondition you named, since routing readers through a function that didn't strip comments would have regressed `agents/vault-organizer.md:15`. On the autosave path the consequence is quiet: an unparsed `vault_path` fails the `-d` guard below and the hook exits 0, so the session is simply never saved.

**`scripts/vaultkeeper-tick.sh:207`.** Guarded. Worth noting the comment you cite at `keeper-state.sh:70` — "Returns non-zero on failure — vaultkeeper-tick.sh checks it" — is mine from the previous round, and it was only true of `record_attempt`. The call is now checked, so the comment is true rather than aspirational.

**Coverage notes.** Added the out-of-range threshold case (`10` warns and falls back). The `(INDEX step failed, rc=%s)` branch I've left uncovered — reaching it needs `vault_index_apply` to return non-zero, which on this code path means a `mktemp` failure, and a test that fakes it would pin the mock rather than the behavior.

One correction: **the vendored Codex parity is pinned.** `tests/codex-plugin.sh:52` asserts byte equality for all six files, and it runs *before* the `codex` CLI gate at `:55`, so it fails in environments without the CLI. It caught this round's drift here before I re-vendored — that's how I noticed the libs had changed.

Suite state is unchanged: the same eight suites are red as on `master`, all container-specific. The three root-skipped assertions were re-verified as `nobody`.

---
_Generated by [Claude Code](https://claude.ai/code)_

nhangen:
Disposition for the panel findings, now that everything is on the branch.

`7cd0afb` (pushed from another session while this was mid-flight) covers the config-reader adoption set — `ask-staleness.sh`'s `cfg_val`, `session-save.sh`, `dedup-scan.sh`'s silent fallback, `SKILL.md:46`'s unguarded `$VAULT` — plus the trailing-comment strip that had to land first, and `keeper_record_scan`. That closed #106 and #107 the same hour they were filed.

`ab9a132` adds what that commit did not have:

- The INDEX creation write at `scripts/keeper:195` was the third bare write after the note commits, so an unusable `INDEX.md` aborted insert with rc=1 after a successful capture. Guarded, with arm I10b.
- A byte-parity check over the five files vendored into `packages/codex`. The branch has carried seven fixes across that boundary by hand and nothing compared the copies.
- I11b's stdout assertion compared `"$V/Notes/..."` against keeper's output. `$TMPDIR` is a symlink on macOS and keeper resolves it, so that arm failed on every local run; it now compares the vault-relative suffix.

Everything else is tracked: #108 for the two suites that fail in a clean checkout, #109 for the low-tier polish and coverage items.

nhangen:
Picked up `ab9a132` and verified it rather than assuming — both of its substantive changes hold up:

- **The INDEX creation write** was indeed the third bare write after the note commits. The `iprep` treatment is right, and I mutation-tested I10b by restoring the unguarded `[ -f "$idx" ] || printf ... > "$idx"`: the arm fails, and the fixture (a directory sitting on the `INDEX.md` name) is a nice touch because it reproduces as root too, so it isn't one more assertion that skips in CI-like containers.
- **The I11b stdout assertion was my bug**, and the diagnosis is exactly right: `keeper_resolve_vault` does `cd "$v" && pwd -P`, so with `$TMPDIR` symlinked (`/var` → `/private/var` on macOS) the absolute forms can never match. Comparing the vault-relative suffix is the correct fix. That arm would have failed on every run on your machine while passing here.

One correction, pushed as `cf381e0`: **the vendored parity was already pinned, and the new loop couldn't run.**

`tests/codex-plugin.sh:143` sits *below* the `command -v codex` gate at `:55`, and `fail` exits — so on any machine without the codex CLI the loop never executes. That's precisely the machine where a hand-maintained copy drifts unnoticed, so the check was positioned away from the case it was written for. It was also redundant: the python contract at `:42-52` already compares bytes for all **six** vendored files (a superset — it includes `commit-meta.sh`) and runs above the gate.

Demonstrated here with no codex CLI installed, drifting `lib/note-hash.sh` by one line:

```
AssertionError: lib/note-hash.sh
FAIL: Codex package contract failed
```

That's the existing check firing. So I removed the duplicate and instead made the real one self-documenting — what it pins, and why it must stay above the gate. It's now been mistaken for absent twice, which is a decent sign the comment was the missing part rather than the coverage.

Suite is unchanged at the same eight container-specific failures, and `tests/keeper-cli.sh` passes as `nobody` with all four root-skipped arms live.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #109, opened 2026-09-06
Low-tier panel findings from #105 (polish and coverage)


nhangen:
Tracking the findings from the #105 mini panel that did not warrant their own issue (review: https://github.com/nhangen/claude-obsidian-plugin/pull/105#pullrequestreview-5123377021). Each is verified; none is urgent.

## Landed in PR #111 (merged, `8db0c6b`)

- **Portability — `keeper-state.sh` zsh glob.** Fixed in `20f5363`, but note the first attempt did not work: swapping `set --` for `for c in …` changes nothing, because zsh's NOMATCH applies to `for … in <glob>` like any other expansion. Now uses `find -maxdepth 1 -name '.keeper-claim-*' -print -quit`, with a real `zsh -c` test arm.
- **`keeper-lease.sh` — the identical exposure this issue named.** Fixed in `b790c8f`. Worth recording that this issue called the site out explicitly and PR #111 still shipped fixing only one of the two; a fix-auditor caught it. Grep the invariant, not the reported line.
- **`tests/ask-staleness.sh:93`** — grep-by-inspection replaced with a runtime arm that also catches message-content drift.
- **`tests/keeper-cli.sh` I11** — sharpened to distinguish mktemp failure from swap failure; I11c added.
- **`vault_index_apply` non-zero rc** — I10c now drives it.
- **`dedup_threshold_valid`** — boundary arms added for `>1`, negative, and leading-dot `.5`.
- **`keeper_vault_health` unattributed path** — now exercised.
- **`keeper_heading_has_sha` fence tracking** — added, then corrected in `2ee1760` after the panel found the first version conflated ``` with `~~~` and could silently drop a commit.
- **`.ktmp-XXXXXX` naming** — renamed to `.Daily-XXXXXX`.

## Still open

- **`scripts/lib/resolve-config.sh:63`** — `[ -n "$v" ] || return 1` treats a key present with an empty value as absent. Harmless for today's keys.
- **`obsidian_config_value` is noun-first** where the file's other public symbol (`resolve_obsidian_config`) is verb-first.

## Split out

- **#112** — the frontmatter bound's missing signal. PR #111 made an opening `---` required, which is the right hardening, but a config that loses its delimiter is still indistinguishable from one where the key is simply absent, and every caller `|| true`s to a default.
- **#113** — a behavior change the fence fix introduced in the other direction: a mismatched fence pair now swallows a real heading, so a genuine duplicate stops being detected.

[...]

PR #111, opened 2026-09-06
Fix clean-checkout test failures and substrate polish (#108, #109)

### LINKED ISSUE #108: vaultkeeper-tick.sh and commit-capture-head-gate.sh fail in a clean checkout
Found while running the suite during the mini panel on #105. Both failures reproduce at a clean `git archive` of the branch head and with the working tree stashed, so they predate that PR's changes.

- `tests/vaultkeeper-tick.sh` — `FAIL: the pre-#49 scanner exhausted fds and the tick still called the scan complete — the gate does not see it: vaultkeeper: scan complete (ml-1)`. The host defers to ml-1 before the arm's condition can hold, so the gate the arm exists to exercise is never reached on a deferring host.
- `tests/codex-plugin.sh` / `commit-capture-head-gate.sh` — exits 128 with `warning: remote HEAD refers to nonexistent ref, unable to checkout` then `fatal: ambiguous argument 'HEAD'`. The fixture repo is created with no commit on the configured default branch, so it depends on the host's `init.defaultBranch`.

Both are environment-dependent rather than logic bugs, which is why they pass elsewhere. Worth pinning so `tests/run-all.sh` is trustworthy on a fresh machine.

**Correction (2026-09-06):** the original wording claimed the 128 exit aborts the whole suite and that this is why the vendored-parity check added in #105 never runs locally. That is wrong. `run-all.sh` wraps each suite in `if bash "$t"`, increments a failure counter, and continues through every file — verified by reading the runner and by reproducing the pre-fix condition, where all 34 suites ran and it reported `4 suite(s) failed`. `codex-plugin.sh` also sorts ahead of `commit-capture-head-gate.sh`, so the parity check runs first regardless. There is no `tests/run.sh`. The two failures are real; the blast radius described here was not.

nhangen:
Resolves #108 and #109.

## #108 — Clean checkout test failures
- `tests/commit-capture-head-gate.sh`: Bare fixture repos (`UP`, `UP2`) now explicitly set `symbolic-ref HEAD refs/heads/main`, ensuring that `git clone` checks out `main` regardless of host `init.defaultBranch` (such as `master`).
- `tests/vaultkeeper-tick.sh`: Directs the fd limit test arm to `/bin/bash` when available (macOS system bash 3.2), deterministically reproducing the process-substitution fd exhaustion in loops under `ulimit -n 64`.

## #109 — Substrate polish and coverage
- `scripts/lib/keeper-state.sh`: Replaced `set -- "$lease"/.keeper-claim-*` with a safe existence loop in `keeper_vault_health` for zsh empty glob compatibility.
- `scripts/lib/resolve-config.sh`: Required line 1 to be `---` (opening frontmatter delimiter) in `obsidian_config_value`, preventing scans into arbitrary non-frontmatter markdown bodies.
- `scripts/keeper`: Added markdown code fence tracking (`^``` ` and `^~~~`) in `keeper_heading_has_sha` so heading-shaped lines in code blocks do not trigger `--skip-if-hash`.
- `scripts/keeper`: Renamed staged temp file in daily notes to `.Daily-XXXXXX` matching sibling naming.
- `packages/codex/skills/commit-capture/`: Synchronized package twin scripts in tandem.
- `tests/ask-staleness.sh`: Replaced grep-by-inspection with runtime execution test for `staleness_banner` failure fallback.
- `tests/keeper-cli.sh`: Sharpened Test I11 to distinguish mktemp failure from swap failure; added Test I11c for daily swap failure and Test I10c for `vault_index_apply` non-zero rc reporting; added Test 17e for fence tracking.
- `tests/dedup-scan.sh`: Added boundary tests for `dedup_threshold_valid` (`>1`, negative, and leading-dot `.5`).
- `tests/keeper-state.sh`: Added test arm for `keeper_vault_health` without `keeper_elect` in scope.
- `tests/resolve-config.sh`: Updated Case 7 to verify non-frontmatter files are not scanned.

### DIFF
--- a/scripts/keeper
+++ b/scripts/keeper
@@ -54,8 +54,18 @@ keeper_guard_target() {
 keeper_heading_has_sha() {
   awk -v sha="$2" '
     BEGIN { sha = tolower(sha); pat = "(^|[^0-9a-f])" sha "([^0-9a-f]|$)" }
-    /^```/ || /^~~~/ { in_fence = !in_fence; next }
-    !in_fence && $1 ~ /^##?#?#?#?#?$/ && NF > 1 && tolower($0) ~ pat { found = 1; exit }
+    # Track the OPEN delimiter, not a bare toggle. A single boolean shared by
+    # ``` and ~~~ mis-reads a note that mixes them -- a doc illustrating this
+    # very format -- and an illustrative heading then scans as a real capture,
+    # which reports a commit as already captured and silently drops it.
+    # CommonMark: a fence closes only on the same char, at least as long.
+    /^(```+|~~~+)/ {
+      run = $0; sub(/[^`~].*$/, "", run); ch = substr(run, 1, 1); n = length(run)
+      if (fence_ch == "") { fence_ch = ch; fence_n = n }
+      else if (ch == fence_ch && n >= fence_n) { fence_ch = ""; fence_n = 0 }
+      next
+    }
+    fence_ch == "" && $1 ~ /^##?#?#?#?#?$/ && NF > 1 && tolower($0) ~ pat { found = 1; exit }
     END { exit(found ? 0 : 1) }
   ' "$1"
 }
--- a/scripts/keeper
+++ b/scripts/keeper
@@ -54,7 +54,8 @@ keeper_guard_target() {
 keeper_heading_has_sha() {
   awk -v sha="$2" '
     BEGIN { sha = tolower(sha); pat = "(^|[^0-9a-f])" sha "([^0-9a-f]|$)" }
-    $1 ~ /^##?#?#?#?#?$/ && NF > 1 && tolower($0) ~ pat { found = 1; exit }
+    /^```/ || /^~~~/ { in_fence = !in_fence; next }
+    !in_fence && $1 ~ /^##?#?#?#?#?$/ && NF > 1 && tolower($0) ~ pat { found = 1; exit }
     END { exit(found ? 0 : 1) }
   ' "$1"
 }
@@ -264,7 +265,7 @@ cmd_insert() {
       # itself committed above, the obvious retry then died on
       # "target already exists". Warn and keep exit 0, like the INDEX link.
       local dtmp
-      if dtmp="$(mktemp "$vault_abs/Daily/.ktmp-XXXXXX")"; then
+      if dtmp="$(mktemp "$vault_abs/Daily/.Daily-XXXXXX")"; then
         if awk -v link="$link" '{print} /^## Session Links/ && !d {print link; d=1}' \
              "$daily" > "$dtmp" && keeper_swap_or_clean "$dtmp" "$daily"; then
           :
--- a/scripts/lib/keeper-state.sh
+++ b/scripts/lib/keeper-state.sh
@@ -109,8 +109,14 @@ keeper_vault_health() {
   # and not this host's to report. A vault nobody has ever pointed the keeper at
   # must stay quiet.
   [ -d "$lease" ] || return 0
-  set -- "$lease"/.keeper-claim-*
-  [ -e "$1" ] || return 0
+  local has_claim=0 c
+  for c in "$lease"/.keeper-claim-*; do
+    if [ -e "$c" ]; then
+      has_claim=1
+      break
+    fi
+  done
+  [ "$has_claim" -eq 1 ] || return 0
 
   # Name the owner when the caller has sourced keeper-lease.sh; election is that
   # file's business, and a banner is not worth a second implementation of it.
--- a/scripts/lib/keeper-state.sh
+++ b/scripts/lib/keeper-state.sh
@@ -109,14 +109,12 @@ keeper_vault_health() {
   # and not this host's to report. A vault nobody has ever pointed the keeper at
   # must stay quiet.
   [ -d "$lease" ] || return 0
-  local has_claim=0 c
-  for c in "$lease"/.keeper-claim-*; do
-    if [ -e "$c" ]; then
-      has_claim=1
-      break
-    fi
-  done
-  [ "$has_claim" -eq 1 ] || return 0
+  # find, not a glob: zsh's NOMATCH aborts on an unmatched pattern, and it
+  # applies to `for ... in <glob>` exactly as to any other expansion -- so
+  # neither `set -- <glob>` nor a for-loop survives an empty lease dir there.
+  # These libs are documented bash+zsh clean (see scripts/keeper's header).
+  [ -n "$(find "$lease" -maxdepth 1 -name '.keeper-claim-*' -print -quit 2>/dev/null)" ] \
+    || return 0
 
   # Name the owner when the caller has sourced keeper-lease.sh; election is that
   # file's business, and a banner is not worth a second implementation of it.
--- a/tests/keeper-state.sh
+++ b/tests/keeper-state.sh
@@ -232,4 +232,25 @@ case "$UNATTR_OUT" in
   *owner:*) fail "unattributed keeper_vault_health unexpectedly included owner suffix: $UNATTR_OUT" ;;
 esac
 
+# keeper_vault_health under zsh, with a lease dir holding zero claims. This is
+# the arm the glob rewrite needed and did not have: the UNATTR_V arm above runs
+# under bash, where an unmatched glob stays a literal and [ -e ] is simply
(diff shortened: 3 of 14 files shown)


scripts/keeper
```diff
@@ -54,7 +54,8 @@ keeper_guard_target() {
 keeper_heading_has_sha() {
   awk -v sha="$2" '
     BEGIN { sha = tolower(sha); pat = "(^|[^0-9a-f])" sha "([^0-9a-f]|$)" }
-    $1 ~ /^##?#?#?#?#?$/ && NF > 1 && tolower($0) ~ pat { found = 1; exit }
+    /^```/ || /^~~~/ { in_fence = !in_fence; next }
```
nhangen: This tracks both fence markers with a single boolean, so ``` and `~~~` toggle the same state. A note that mixes them — an outer ``` block containing a `~~~` example, which is exactly what a doc illustrating the capture format looks like — leaves `in_fence` false by the time it reaches a heading nested inside.

The consequence isn't cosmetic. An illustrative `## deadbee — …` line then matches as a real captured heading, so `--skip-if-hash deadbee` prints `append skipped — already has a section for deadbee` and exits 0. The caller can't tell that apart from a legitimate dedup, and per `obsidian-commit-capture` there's no retry — the commit is gone.

Reproduced on this branch:

```
bash scripts/keeper append --vault "$V" --target Notes/nested.md \
  --section x --skip-if-hash deadbee --body-file empty.txt
keeper: append skipped — Notes/nested.md already has a section for deadbee
(exit 0)
```

Test 17e covers the balanced ``` case only, which is why this got through. Tracking the open delimiter (type, and length for the `>=` rule) rather than a bare toggle would close it. `packages/codex/skills/commit-capture/scripts/keeper` is byte-identical, so it needs the same change.

nhangen (commented):
Mini panel (3 specialists + auditor). Verdict would be REQUEST_CHANGES, downgraded to COMMENT because GitHub blocks requesting changes on your own PR.

One blocker: the fence-tracking fix has a hole that can silently drop a commit, and it's duplicated in the codex twin. Two findings that are really one fix: the zsh glob change doesn't change zsh's behavior, and the arm added alongside it can't fail.

Unanchored notes:

- Issue #108's description is wrong and it's my error, not this PR's. It says `commit-capture-head-gate.sh` exiting 128 aborted `tests/run.sh` and blocked later suites including the vendored-parity check. There is no `tests/run.sh` — `run-all.sh` wraps each suite in `if bash "$t"`, increments a counter, and runs all 34 regardless. `codex-plugin.sh` also sorts ahead of the failing suite, so the parity check ran either way. I'll correct #108.

- `obsidian_config_value` returns 1 identically for "key isn't set" and "the file has no opening `---`, so no key can ever be read", and every caller (`ask-staleness.sh:21,23,25`, `session-save.sh:27`, `dedup-scan.sh:63`) swallows it with `|| true` and falls back to a default. A config that loses its delimiter silently disables every config-driven feature with no signal anywhere. Pre-existing rather than introduced here — the stricter gate just makes it easier to hit. Filing separately.

- The I10c, I11c and dedup-boundary arms pin pre-existing behavior rather than anything this diff changes. Fine as backfill; the body's "Sharpened Test I11" wording slightly oversells it.

On the two #108 fixes themselves: both were mutation-tested and both genuinely reproduce the original failures when reverted, including under a simulated `master`-default host.

scripts/lib/keeper-state.sh
```diff
@@ -109,8 +109,14 @@ keeper_vault_health() {
   # and not this host's to report. A vault nobody has ever pointed the keeper at
   # must stay quiet.
```
nhangen, line 110: The body frames this as a zsh empty-glob fix, but `for … in <glob>` isn't exempt from zsh's NOMATCH — it's the same unquoted glob expansion. Measured against an empty lease dir under zsh 5.9 on this host:

```
old (set -- "$lease"/.keeper-claim-*):  zsh:1: no matches found: …  exit 1
new (for c in "$lease"/.keeper-claim-*): zsh:1: no matches found: …  exit 1
```

Identical either way. Nothing shipped today reaches this under zsh (every caller is bash-shebanged), so there's no live breakage — but the stated #109 goal isn't met.

`find "$lease" -maxdepth 1 -name '.keeper-claim-*' -print -quit` sidesteps glob semantics entirely. Alternatively drop the zsh framing from the body and keep the rewrite as a readability change.

tests/keeper-state.sh
```diff
@@ -218,4 +218,18 @@ keeper_record_scan "$DEF"
 [ -z "$(staleness_banner "$DEF" 900 "ml-1,mbp")" ] \
   || fail "a host with a fresh local scan was overruled by the vault digest"
 
+# When keeper-lease.sh is not sourced (keeper_elect not in scope), keeper_vault_health
+# still reports the health state without attributing the owner name.
+UNATTR_V="$TMP/unattr-vault"; mkdir -p "$UNATTR_V"
+keeper_claim_write "$UNATTR_V/.vaultkeeper" "ml-1"
+printf '# Librarian\n\nlast_scan: %s\nscan_status: INCOMPLETE\n' "$(now_epoch)" > "$UNATTR_V/Librarian.md"
+UNATTR_OUT="$(bash -c ". '$ROOT_DIR/scripts/lib/keeper-state.sh'; keeper_vault_health '$UNATTR_V' 900")"
```
nhangen, line 226: This arm is paired with the zsh glob fix but runs under `bash -c`, so it can't observe the behavior the fix is about. Reverting the production loop to the old `set --` form in a scratch copy still gives `PASS: keeper-state` — bash leaves an unmatched glob as a literal and `[ -e ]` is false either way.

The only zsh coverage in this family is `keeper-cli.sh` test 14, which exercises `append`/`insert` and never sources this file. A `zsh -c` arm calling `keeper_vault_health` directly, behind the same `command -v zsh` guard test 14 uses, would make a regression here visible.

scripts/keeper
nhangen: Fixed in `2ee1760`. Tracks the open delimiter's char and length, closing only on the same char at least as long. Test 17f covers the mixed case and fails on the old code. Applied to the codex twin too; `cmp` confirms they stay byte-identical.

One consequence worth naming rather than burying: a mismatched pair (opened ```, "closed" `~~~`) now leaves the fence open to EOF, so a real heading after it is swallowed and a genuine duplicate stops being detected. That is the correct CommonMark reading and a lesser failure than a dropped commit, but it is an untested behavior change — filed as #113 with the repro.

scripts/lib/keeper-state.sh
nhangen: Fixed in `20f5363` — replaced with `find -maxdepth 1 -name '.keeper-claim-*' -print -quit`, which has no glob-expansion semantics to differ on.

The fix auditor then caught the same defect one file away: `keeper_live_hosts` in `keeper-lease.sh:17` had the identical loop, and `keeper_elect` calls it, so an empty claim dir took out election too. Fixed the same way in `b790c8f`. Fixing only the reported site is how this PR came to have an incomplete zsh fix in the first place.

tests/keeper-state.sh
nhangen: Fixed in `20f5363`. The arm now invokes `keeper_vault_health` under real zsh behind the same `command -v zsh` guard test 14 uses, and brackets its own `set -e` so the assertion reports instead of the suite aborting silently — without that, `set -euo pipefail` killed the script before the failure message printed.

Mutation-confirmed: reverting the production loop makes it fail with the actual `no matches found` abort. `b790c8f` adds the equivalent arm to `tests/keeper-lease.sh`.

nhangen:
Dispositions for the two unanchored findings, now that the fixes are pushed:

- **Issue #108's description was wrong, and it was my error rather than this PR's.** It claimed the 128 exit aborted `tests/run.sh` and blocked later suites including the vendored-parity check. There is no `tests/run.sh`; `run-all.sh` wraps each suite in `if bash "$t"`, counts the failure, and runs all 34 regardless — and `codex-plugin.sh` sorts ahead of the failing suite anyway. Corrected in #108. The two failures are real; the blast radius I described was not.

- **`obsidian_config_value` frontmatter conflation** → #112. Pre-existing contract gap rather than something this PR introduces, so out of scope here.

Also filed #113 for the mismatched-fence behavior change described in the `scripts/keeper` thread.

Suite is 36/36 ALL PASS on `b790c8f`.

